PR #123, opened 2026-06-12
fix(unmarshal): skip non-participating optional groups like Decoder

### LINKED ISSUE #104: Bug: Unmarshal errors on non-participating optional groups; Decoder silently skips them
The two decode paths disagree on the most common edge case in this library: an optional named group that did not participate in the match.

**Repro:** pattern `(?P<name>\w+)(?: (?P<age>\d+))?` against `"Alice"`, struct with `Age int`:
- `Unmarshal` returns `regextra: failed to set field Age: cannot convert "" to int`
- `Decoder.One` with the identical pattern/struct **succeeds**, leaving `Age` zero

**Cause:** in `populateStruct` (unmarshal.go:205-217), the empty-value guard only substitutes a `default=` value; without a default, `found` stays true and `""` is fed to `strconv.ParseInt`. The decoder path (decoder.go:282-296) treats `""` as not-found and skips the field.

**Fix:** in `populateStruct`, `continue` when `value == ""` and no default is set — mirroring `decode()`. The Decoder semantic (skip, leave zero value) is the right one for optional groups.

Neither behavior is currently documented or tested. This should be designed together with the planned `regex:"name,required"` tag flag (ROADMAP), since both answer "what does a missing group mean."

Priority: **P0** — optional groups are a bread-and-butter use case; two independent reviews flagged this.

Jecoms:
Fixes #104.

`Unmarshal` and `Decoder.One` disagreed on the most common edge case in the library: an optional named group that didn't participate in the match. With `(?P<name>\w+)(?: (?P<age>\d+))?` against `"Alice"`, `Unmarshal` returned `cannot convert "" to int` while `Decoder.One` succeeded with `Age` left zero.

populateStruct's empty-value guard only substituted a `default=`; with no default it fed `""` to the type converter. It now skips the field entirely (mirroring `decode()`): a group that didn't participate, or matched an empty span, leaves the field unchanged. The Decoder semantic wins — optional groups are data absence, not a conversion failure.

Documented in the `Unmarshal` godoc's field-mapping rules. Tests: new `unmarshal_optional_test.go` including an Unmarshal/Decoder parity test so the two paths can't drift again on this case. The full existing suite passes unchanged — nothing pinned the old erroring behavior.


Jecoms:
## Code review

**Verdict:** The core change is correct — Unmarshal now skips non-participating/empty groups exactly like `Decoder.decode` (parity test included; full suite passes). But it ships a second observable behavior change that the repo's own stability contract flags as breaking, with no CHANGELOG note, and the skip semantics are now duplicated across two files.

### Findings

**1. [medium] Second silent behavior change: pre-populated string fields are no longer reset** — `unmarshal.go:211`
Previously, a declared group that didn't participate (or matched empty) wrote `""` into string fields, overwriting pre-populated values; now the field is left untouched. Scenario: a caller declares `var p struct{ Name, Note string }` once outside a loop and calls Unmarshal per log line, relying on an absent optional `note` group resetting `Note` to `""` — after upgrading, record 2 silently carries record 1's `Note` (cross-record data bleed). It's intentional and tested, and consistent with Decoder for *fresh* structs — but Decoder always starts from a zero `T`, so "mirroring Decoder" doesn't actually entail retaining stale caller state. README §Stability lists "changing observable behavior of an existing call" as breaking; at minimum this needs a release-notes callout.

**2. [low] Skip-or-default semantics now duplicated, and the Decoder copy is convoluted** — `decoder.go:286` vs `unmarshal.go:211-217`
The two paths implement the same contract with different shapes — and this PR exists precisely because they drifted (issue #104). The Decoder copy also has a dead clause: in `found = value != "" || fd.groupIndex == 0`, `groupIndex` can never be 0 (compileDecoder only stores `SubexpIndex` of a non-empty group name), and the follow-up `if !found || value == ""` re-tests `value == ""` redundantly. A shared helper (e.g. `resolveGroupValue(value, found, opts) (string, bool)`) would lock the contract together.

**3. [low] No CHANGELOG entry** — `CHANGELOG.md`
A user-visible behavior fix with no entry under `[Unreleased]`, in a repo whose CHANGELOG records per-PR links for every prior behavior change.

**4. [low] Test gap: participating empty-span group without a default** — `unmarshal_optional_test.go:13`
The new doc bullet promises "or that matched an empty span, leaves the field unchanged", but no test covers a group that *participates* with a zero-length match and has no default — the new tests only exercise non-participation, and the only existing empty-span test goes through the default path. A future refactor reintroducing `""` writes (or conversion errors) for that case would pass the whole suite.

**5. [low] Stale doc bullet two lines above the edit** — `unmarshal.go:55`
"Supports type conversion for int, int64, float64, and bool" understates actual support (`setFieldValue` handles all int/uint widths, float32, `time.Time`, `time.Duration`, pointer fields, `RegexUnmarshaler`).

---

Jecoms:

**Verdict: APPROVE WITH NITS.** The branch has been **rebased onto current `main`** (it predated #129/#127) and force-pushed. One `decoder.go` conflict was resolved by **keeping #127's `groupIndexes` occurrence-iteration** and applying only this PR's intent (the shared `resolveGroupValue` helper); the CHANGELOG was reframed to build on #127.

### Perf: maintained
#129's `UnmarshalAll` optimizations are fully intact on the rebased tree — `re.SubexpNames()` hoisted once, pre-sized `reflect.MakeSlice`, reused `clear()`-ed `groupValues` map, decode-in-place via `newSlice.Index(idx)`. The PR doesn't touch that loop. `BenchmarkUnmarshalAll -benchmem`: **allocs/op byte-for-byte identical** on every sub-benchmark (p=1.000), B/op flat. `resolveGroupValue` is a pure, zero-alloc function called once per field exactly as the inline code was.

### Findings
- **[boulder] v1 stability decision (maintainer call, not a code fix).** This introduces a documented observable behavior change: empty-span/absent groups now *skip* string fields instead of overwriting with `""`. The typed-field part (error→nil) is strictly safer; the **string-overwrite→skip** part can return stale data to callers reusing a struct across `Unmarshal` calls. The README lists "changing observable behavior of an existing call" as breaking, with no bug-fix exception, and v1.0.0 just shipped. Decide: accept as bug-fix, or defer to v2.
- **[sand]** Scope now overlaps #127, which already fixed the pure non-participating case. The residual value here is (a) the *participating empty-span* case and (b) the shared `resolveGroupValue` helper preventing future drift. Worth reframing the PR description post-#127.
- **[sand]** `unmarshal_optional_test.go` is internal `package regextra` but uses only the public API; #70 moved tests to external `package regextra_test`. (Precedent is mixed — `duplicate_groups_test.go` is also internal — so non-blocking.)
- **[dust]** `TestUnmarshal_decoderParityOnOptionalGroups` exercises only `"Alice"`/`"Alice 30"`, not the empty-span input that is this PR's actual novel behavior — a small coverage gap for the case being fixed.
- **[dust]** `resolveGroupValue`'s `found` param is redundant (`found && value != ""` ≡ `value != ""` given callers set `value=""` when `!found`). Harmless/self-documenting.

### Tests
`go build`, `go vet`, `go test ./...` all green on the rebased branch. The "participating empty-span leaves field unchanged" case genuinely fails on current `main` (main feeds `""` to the int converter), confirming residual value beyond #127.

<sub>Perf-focused review; branch rebased onto current main and force-pushed (`--force-with-lease`). 🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

Jecoms:
Addressed review feedback in 468a36e.

After the rebase onto current main, the first review's findings were already resolved in the tree (CHANGELOG behavior-change note, shared `resolveGroupValue` helper, removed dead `groupIndex == 0` clause, updated doc bullet). The one remaining gap was test coverage for the PR's actual novel case — a *participating zero-length span* — so I added a parity check on `(?P<name>\w+):(?P<age>\d*)` against `"Alice:"`, asserting `Unmarshal` and `Decoder.One` resolve identically. Test-only; build/vet/test green.

Still gating merge (maintainer call): the string-field overwrite→skip change is observably breaking for struct-reuse callers, and README §Stability lists no bug-fix exception post-1.0.0. The behavior is intended and tested — so the open question is accept-as-patch/minor vs defer-to-v2.

Issue #114, opened 2026-06-12
Tests: cover Replace's documented behaviors


Jecoms:
`Replace` (90.6% coverage) documents three behaviors that have zero test coverage:

1. Non-participating optional groups are skipped (regextra.go:357)
2. Overlapping/nested named groups — "outermost wins" (regextra.go:364); see also #107, which fixes the unstable sort that makes this incidental rather than guaranteed
3. Unnamed groups mixed with named groups (regextra.go:349)

Add table tests for all three.

## Implementation plan

_Added 2026-06-30 via /gh-analyze-issue. Review before implementing._

**Approach.** Add intentional cases to the existing `TestReplace` table in
`regextra_test.go` for the documented behaviors that still lack their own
assertions. The line refs in the issue are stale (they now point at
`NamedGroupsPerMatch`); the engine is `replaceNamed` (`regextra.go:556`).
Current state has shifted since filing (now 100% on `Replace`, 97% on
`replaceNamed`), so only two of the three behaviors actually need new tests:

- **#2 "outermost wins" — already covered.** Two cases ("nested groups,
  outermost wins" / "nested groups, only inner group in map is substituted")
  were added with #107. No new test needed; just confirm they assert the doc.
- **#1 non-participating optional group skipped** (`regextra.go:579`,
  `s < 0 || e < 0`) — branch is only hit *incidentally* by a `TestReplaceFunc`
  case; there is no intentional `Replace` case. Add one.
- **#3 unnamed groups mixed with named** (`regextra.go:575`, `name == ""`) —
  **zero** coverage. Add one.

**Steps.**
1. Add a `TestReplace` case for a non-participating optional named group
   (e.g. pattern with `(?P<opt>...)?` that is absent from the match) — assert
   the rest still substitutes and the absent group is silently skipped.
2. Add a `TestReplace` case mixing an unnamed group with named groups, e.g.
   `(\w+)=(?P<val>\d+)` — assert the unnamed group is left untouched and only
   the named group is replaced.
3. Run `go test -run TestReplace -coverprofile` and confirm `regextra.go:575`
   and `:579` branches are now covered by `TestReplace` specifically.

**Files / surfaces touched.**
- `regextra_test.go` — new cases in the `TestReplace` table (no production code).

**Verification.**
- `go test ./...` passes.
- `go test -run TestReplace -coverprofile=cov.out && go tool cover -func=cov.out`
  shows `replaceNamed` at 100%, with the `name == ""` and `s < 0` skip branches
[...]

PR #124, opened 2026-06-12
fix: enforce outermost-wins deterministically in Replace

### LINKED ISSUE #107: Bug: Replace uses unstable sort — documented "outermost wins" tie-break not guaranteed
`Replace` (regextra.go:362) sorts replacement spans with `sort.Slice` (unstable), ordered only by start offset. For nested groups that start at the same offset — e.g. `(?P<outer>(?P<inner>x)y)` with both names in the replacements map — which span wins depends on sort internals, not the documented "outermost wins" rule.

It currently happens to work for small inputs, but nothing guarantees it.

**Fix:** use `sort.SliceStable`, or better, tie-break explicitly by descending span length so the invariant is enforced rather than incidental.

**Tests:** the doc comment's claims are all currently untested — add cases for same-start nested groups (outermost wins), non-participating optional groups skipped (regextra.go:357), and unnamed groups mixed with named ones (regextra.go:349). (Test additions also tracked in the Replace test-coverage issue.)

Jecoms:
Fixes #107.

`Replace` sorted replacement spans with the unstable `sort.Slice`, ordered only by start offset. For nested groups starting at the same offset — `(?P<outer>(?P<inner>x)y)` with both names in the replacements map — the documented "outermost wins" rule held only by sort-internals luck. Now `sort.SliceStable` with an explicit tie-break (same start → longer span first) enforces the invariant.

Tests: new `replace_order_test.go` covering same-start nesting at two and three levels (where unstable-sort behavior is least predictable), inner-only and outer-only replacement, and non-same-start nesting. These also close part of the untested-documented-behaviors gap tracked in #114.

### DIFF
--- a/regextra.go
+++ b/regextra.go
@@ -359,7 +359,15 @@ func Replace(re *regexp.Regexp, target string, replacements map[string]string) s
 			}
 			spans = append(spans, span{start: s, end: e, repl: repl})
 		}
-		sort.Slice(spans, func(i, j int) bool { return spans[i].start < spans[j].start })
+		// Order by start; when spans start at the same offset (nested groups),
+		// the longer span sorts first so the documented "outermost wins" rule
+		// is enforced rather than left to sort-internals luck.
+		sort.SliceStable(spans, func(i, j int) bool {
+			if spans[i].start != spans[j].start {
+				return spans[i].start < spans[j].start
+			}
+			return spans[i].end > spans[j].end
+		})
 		for _, sp := range spans {
 			if sp.start < cursor {
 				continue // already covered (overlap with an earlier substitution)
--- a/replace_order_test.go
+++ b/replace_order_test.go
@@ -1,0 +1,73 @@
+package regextra
+
+import (
+	"regexp"
+	"testing"
+)
+
+// Regression tests for https://github.com/Jecoms/regextra/issues/107:
+// when nested named groups start at the same offset, the documented
+// "outermost wins" rule must hold deterministically, not depend on
+// sort internals.
+
+func TestReplace_nestedSameStartOutermostWins(t *testing.T) {
+	re := regexp.MustCompile(`(?P<outer>(?P<inner>x)y)`)
+
+	t.Run("both groups in map: outer wins", func(t *testing.T) {
+		got := Replace(re, "axyb", map[string]string{
+			"outer": "OUTER",
+			"inner": "INNER",
+		})
+		if got != "aOUTERb" {
+			t.Errorf("got %q, want %q", got, "aOUTERb")
+		}
+	})
+
+	t.Run("only inner in map: inner replaced", func(t *testing.T) {
+		got := Replace(re, "axyb", map[string]string{"inner": "I"})
+		if got != "aIyb" {
+			t.Errorf("got %q, want %q", got, "aIyb")
+		}
+	})
+
+	t.Run("only outer in map: outer replaced", func(t *testing.T) {
+		got := Replace(re, "axyb", map[string]string{"outer": "O"})
+		if got != "aOb" {
+			t.Errorf("got %q, want %q", got, "aOb")
+		}
+	})
+
+	// Three levels of same-start nesting; outermost must still win even when
+	// the slice has more than two same-start spans (where an unstable sort's
+	// behavior would be least predictable).
+	t.Run("three nested levels", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


Jecoms:
## Code review

**Verdict:** The core change is correct for non-empty spans — the (start asc, end desc) comparator plus `SliceStable` deterministically enforces outermost-wins for nested groups, including exactly-equal spans via subexp-index insertion order. But the new tie-break introduces a regression for zero-width sibling groups, and zero-width *nested* groups still escape the rule entirely.

### Findings

**1. [medium] Regression: zero-width sibling group's replacement is silently dropped** — `regextra.go:369`
The end-descending tie-break sorts a longer group ahead of a zero-width group at the same start offset, then the `sp.start < cursor` guard skips the zero-width one — even though it is adjacent, not nested.
Repro (A/B verified): `(?P<a>x*)(?P<b>y)` on `"-y-"` with `{"a": "[A]", "b": "[B]"}` — `a` matches empty at [1,1), `b` matches [1,2). `origin/main` returns `-[A][B]-`; this PR returns `-[B]-`. The documented "inner groups inside an already-replaced span are not substituted" rule doesn't sanction dropping a participating non-nested group. Possible fix: sort zero-width spans (start==end) before non-empty spans at the same start, or make the skip guard check strict containment rather than `start < cursor`.

**2. [medium] Zero-width *nested* groups still violate outermost-wins** — `regextra.go:372`
The `sp.start < cursor` guard can never fire for empty spans, so both outer and inner replacements get written. Repro (verified on PR head): `(?P<outer>(?P<inner>x*))y` on `"ay"` with `{"outer": "[O]", "inner": "[I]"}` returns `a[O][I]y`, violating the documented rule. Pre-existing on main, but this PR's stated purpose is to enforce exactly that rule deterministically, and this geometry still escapes it.

**3. [low] Test gap: same-start same-end nested spans** — `replace_order_test.go:13`
For `(?P<outer>(?P<inner>x+))` both spans are identical, the comparator reports equivalence both ways, and outer wins only because of `SliceStable` + `SubexpNames` insertion order (verified: returns `-[O]-`). The new tests only cover same-start/different-end. A future refactor swapping `SliceStable` back to `sort.Slice` (plausible, since the comparator now looks like a total order) would silently reintroduce nondeterminism with no test failing.

**4. [low] Per-match allocation** — `regextra.go:346`
`spans := make([]span, 0, len(names))` allocates inside the match loop; hoisting it before the loop and reusing via `spans = spans[:0]` removes one allocation per match. Pre-existing, but the sort rewrite touched this block.

---

Jecoms:

**Verdict: REDUNDANT — recommend close (or repurpose to tests-only).** Not rebased/pushed, pending that decision.

### Relationship to #129: superseded
#129 already landed the identical outermost-wins determinism fix. `main`'s `Replace` uses `slices.SortStableFunc(spans, …)` ordered by **start ascending, then end descending** — byte-for-byte the same tie-break this PR adds via `sort.SliceStable`. #129's version is strictly superior: newer `slices`/`cmp` packages and it reuses the span slice across matches (an allocation win this PR lacks). A trial rebase onto current `main` collapses this PR's entire `regextra.go` change to a **zero-line diff** — and would have re-introduced the now-dropped `sort` import.

### Perf: maintained (provably)
With `regextra.go` byte-identical to main, `Replace` runtime/allocations are unchanged by definition. `BenchmarkReplace -benchmem` reflects #129's numbers (e.g. `matches100` 116 allocs/op, `nestedGroups` 4, `emptyReplacements` 0). Single-group extraction and the O(n log n) stable sort are intact.

### Findings
- **[pebble]** The production fix is dead-on-arrival post-#129 — rebased, it's a no-op.
- **[sand]** New tests deviate from `CONTRIBUTING.md:51` (table-driven); `TestReplace_nestedSameStartOutermostWins` uses repeated `t.Run` blocks rather than table rows.
- **[dust]** Two cases ("outer wins", "only inner") duplicate rows #129 already added to `TestReplace` (`regextra_test.go:362-373`).
- **[dust]** Genuinely net-new coverage *does* exist and is worth keeping: only-outer, three-level same-start nesting, and non-same-start nesting — ideally folded into the existing `TestReplace` table.

### Recommendation
Close as superseded by #129, **or** repurpose down to just the 3 net-new determinism cases folded into the table-driven `TestReplace`. Either way no perf-regression risk — main's hot-path work is fully preserved.

<sub>Perf-focused review; not rebased pending close/repurpose decision. 🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

Jecoms:
closing as solved by #129

Issue #105, opened 2026-06-12
Bug: duplicate group names mishandled in both decode paths


Jecoms:
For patterns with duplicate group names (legal in Go >= 1.22, e.g. `(?:(?P<word>a)|(?P<word>b))`), both paths get it wrong, in opposite directions:

- **`NamedGroups` / `Unmarshal` / `UnmarshalAll`** (regextra.go:251-255, regextra.go:296-298, unmarshal.go:97-102, unmarshal.go:157-163): the name->value map is built by iterating all groups, so a *non-participating later* occurrence clobbers the real value with `""`. Matching `"a"` yields `map[word:""]`.
- **`Decoder`** (via `SubexpIndex`): returns the *first* occurrence only, so matching `"b"` yields a zero `Word`.

Since `AllNamedGroups` exists specifically to support duplicate names, the rest of the API mishandling them is notable.

**Fix:** when building the name->value map, don't overwrite a non-empty value with `""`. Fully correct participation detection requires `FindStringSubmatchIndex` (a group participated iff its index pair is >= 0) — worth using since it also distinguishes "matched empty string" from "did not participate".

Add tests covering duplicate names through `Unmarshal` and `Decoder` (currently only `NamedGroups`/`AllNamedGroups` have duplicate-name tests).

PR #125, opened 2026-06-12
refactor(unmarshal): remove unreachable value-receiver dispatch branch

### LINKED ISSUE #109: Cleanup: dead/unreachable-code review for setFieldValue and decode
Both items resolved — closing via #125.

1. **`setFieldValue` `Implements` fallback** — withdrawn as "dead": it is reachable for **interface-typed** fields (`field.Addr().Interface()` is a `*interface` that does not satisfy `RegexUnmarshaler`, so the `CanAddr` branch cannot cover them). The original "every value reaching `setFieldValue` is addressable, so `*T`'s method set makes the fallback redundant" reasoning holds only for concrete field types. #125 adds a regression test for the interface-field case instead of removing the branch.

2. **`decoder.go` `|| fd.groupIndex == 0` clause** — already removed by #127's decode-loop rewrite (`groupIndex int` → `groupIndexes []int`; `subexpIndexes` skips group 0 via `i != 0`, and non-participation is handled by `start < 0`). No code left to remove on `main`.

Jecoms:
Closes #109.

Every `reflect.Value` reaching `setFieldValue` is addressable (struct fields via a pointer's Elem, `reflect.New(...).Elem()`, or a recursive Elem of a pointer field), and `*T`'s method set includes `T`'s value-receiver methods — so the `CanAddr` branch already dispatches value-receiver `RegexUnmarshaler` implementations. The `Implements` fallback below it was unreachable, and worse, if it ever *were* reached it would call `UnmarshalRegex` on a copy: silent success, nothing written.

Removed the branch and the now-unused `regexUnmarshalerType` package var; expanded the `CanAddr` comment to state the addressability invariant. The existing value-receiver test passes unchanged through the `CanAddr` path, confirming redundancy.

The issue's second item — the dead `|| fd.groupIndex == 0` condition in `decoder.go` — is removed wholesale by the decode-loop rewrite in the #105 PR rather than here, to keep the two PRs conflict-free.

### DIFF
--- a/unmarshal.go
+++ b/unmarshal.go
@@ -315,6 +315,10 @@ type RegexUnmarshaler interface {
 	UnmarshalRegex(value string) error
 }
 
+// regexUnmarshalerType is the reflect.Type of RegexUnmarshaler, cached so
+// the implements-check on every field doesn't pay the reflect.TypeOf cost.
+var regexUnmarshalerType = reflect.TypeOf((*RegexUnmarshaler)(nil)).Elem()
+
 // setFieldValue sets the field value with appropriate type conversion.
 // `opts` carries per-field tag options parsed from `regex:"name,key=value,..."`.
 // Currently consulted: `layout` (for time.Time fields). Pass nil for no opts.
@@ -340,13 +344,23 @@ func setFieldValue(field reflect.Value, value string, opts map[string]string) er
 	//    below. A type `type MyTime time.Time` with its own UnmarshalRegex
 	//    must NOT be pre-empted by the time.Time fast path.
 	//
-	//    Every value reaching this function is addressable (struct fields via
-	//    a pointer's Elem, reflect.New(...).Elem(), or a recursive Elem of a
+	//    Values reaching this function are addressable (struct fields via a
+	//    pointer's Elem, reflect.New(...).Elem(), or a recursive Elem of a
 	//    pointer field), and *T's method set includes T's value-receiver
-	//    methods, so this single check dispatches both pointer-receiver and
-	//    value-receiver implementations.
+	//    methods, so for concrete field types this check dispatches both
+	//    pointer-receiver and value-receiver implementations.
 	if field.CanAddr() {
 		if u, ok := field.Addr().Interface().(RegexUnmarshaler); ok {
+			return u.UnmarshalRegex(value)
+		}
+	}
+	// The CanAddr check above does NOT cover interface-typed fields: an
+	// interface field's Addr() is a *interface, which does not satisfy
+	// RegexUnmarshaler even when the stored concrete value does. Dispatch
+	// those on the field's own type/value (e.g. `F RegexUnmarshaler`
+	// pre-populated with a concrete implementation).
+	if field.Type().Implements(regexUnmarshalerType) {
+		if u, ok := field.Interface().(RegexUnmarshaler); ok {
 			return u.UnmarshalRegex(value)
 		}
 	}
--- a/unmarshal.go
+++ b/unmarshal.go
@@ -315,10 +315,6 @@ type RegexUnmarshaler interface {
 	UnmarshalRegex(value string) error
 }
 
-// regexUnmarshalerType is the reflect.Type of RegexUnmarshaler, cached so
-// the implements-check on every field doesn't pay the reflect.TypeOf cost.
-var regexUnmarshalerType = reflect.TypeOf((*RegexUnmarshaler)(nil)).Elem()
-
 // setFieldValue sets the field value with appropriate type conversion.
 // `opts` carries per-field tag options parsed from `regex:"name,key=value,..."`.
 // Currently consulted: `layout` (for time.Time fields). Pass nil for no opts.
@@ -343,14 +339,14 @@ func setFieldValue(field reflect.Value, value string, opts map[string]string) er
 	//    conversions beat everything, including the stdlib special-cases
 	//    below. A type `type MyTime time.Time` with its own UnmarshalRegex
 	//    must NOT be pre-empted by the time.Time fast path.
+	//
+	//    Every value reaching this function is addressable (struct fields via
+	//    a pointer's Elem, reflect.New(...).Elem(), or a recursive Elem of a
+	//    pointer field), and *T's method set includes T's value-receiver
+	//    methods, so this single check dispatches both pointer-receiver and
+	//    value-receiver implementations.
 	if field.CanAddr() {
 		if u, ok := field.Addr().Interface().(RegexUnmarshaler); ok {
-			return u.UnmarshalRegex(value)
-		}
-	}
-	if field.Type().Implements(regexUnmarshalerType) {
-		// Value-receiver method. Rare but possible.
-		if u, ok := field.Interface().(RegexUnmarshaler); ok {
 			return u.UnmarshalRegex(value)
 		}
 	}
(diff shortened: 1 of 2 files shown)


Jecoms:
## Code review

**Verdict: the unreachability claim is unsound — recommend not merging as-is.** The deleted branch was the only dispatch path for interface-typed struct fields, and its removal is an empirically confirmed regression in the public `Unmarshal` API. (The addressability half of the claim *is* correct: all call sites pass addressable values, so concrete value-receiver types remain covered by the `CanAddr` branch.)

### Findings

**1. [high] Deleted branch is reachable via interface-typed struct fields** — `unmarshal.go:355`
For a field declared as an interface whose method set includes `UnmarshalRegex` (e.g. `F regextra.RegexUnmarshaler`), `field.Addr()` yields a `*interface{...}` which does **not** satisfy `RegexUnmarshaler`, so the surviving `CanAddr` branch never fires. The deleted `field.Type().Implements(regexUnmarshalerType)` + `field.Interface().(RegexUnmarshaler)` branch was the only path that dispatched to the stored implementation.
A/B verified with the same program against `origin/main` vs this PR head:
```go
type holder struct { F rx.RegexUnmarshaler `regex:"f"` }
// pre-populated with &status{}, then:
rx.Unmarshal(re, "open", &h)
```
`origin/main`: returns nil, sets the status via `UnmarshalRegex`. PR head: returns `regextra: failed to set field F: unsupported field type: interface`. Pre-populating an interface-typed field is a supported pattern (Unmarshal documents leaving `*v` unchanged on no-match).

**2. [medium] Replacement comment codifies the incorrect reasoning** — `unmarshal.go:350`
"this single check dispatches both pointer-receiver and value-receiver implementations" is true only for concrete field types; interface-kind fields are not dispatched by the `CanAddr` check. Future readers will trust the comment and build on the false invariant.

**3. [low] Documented dispatch contract not updated** — `docs/v1-readiness.md:201`
The doc still lists dispatch step 3 "RegexUnmarshaler (value-receiver) — rare but possible" as part of the locked-in v1 contract and states "reordering is breaking". This PR removes a documented contract step with no doc/CHANGELOG update — by the doc's own rule, arguably a breaking change landing as a no-op refactor.

**4. [low] Test gap let the regression land green** — `unmarshal_test.go:101`
The only nearby test ("value-receiver implementation is consulted") uses an addressable concrete type, which the `CanAddr` branch handles — so the suite never covered the deleted branch's actually-reachable case (interface-typed field with non-nil value). An interface-typed-field test would guard this permanently.

---

Jecoms:

**Verdict: APPROVE** (one informational `[sand]`; nothing blocking). Branch **rebased onto current `main` (clean, no conflicts)** and force-pushed.

### Is the removed branch truly unreachable? — YES (high confidence)
All 5 `setFieldValue` call sites pass an **addressable** value:
- `Unmarshal` → `rv.Elem()`; `Decoder.One/All/Iter` → `reflect.ValueOf(&x).Elem()`; compile-time probe → `reflect.New(sf.Type).Elem()`; pointer recursion → `field.Elem()`.
- `UnmarshalAll` → `newSlice.Index(idx)` where `newSlice = reflect.MakeSlice(...)` — **#129's decode-in-place path**; MakeSlice elements are addressable, so the invariant still holds post-#129 (re-verified, since #129 changed construction away from `reflect.New().Elem()`+`Append`).

At the removed branch's location `field` is always non-pointer, so `CanAddr()` is always true. The deleted `Implements` branch fired only for value-receiver types `T` — but `*T`'s method set is a superset of `T`'s, so the retained `field.Addr().Interface().(RegexUnmarshaler)` branch already handles those and returns first. Bonus: had it ever been reached it was *buggy* — `field.Interface()` operates on a copy, so a mutating `UnmarshalRegex` would silently no-op. `encoding.TextUnmarshaler` is N/A (appears only in a doc comment; dispatch is solely on the package's own `RegexUnmarshaler`).

### Perf: maintained
#129's work is fully intact on the rebased tree. allocs/op & B/op **identical** main-vs-rebased across all dispatch and hot-path benches (e.g. `Unmarshal/valueReceiverUnmarshaler` 5 allocs/395 B; `UnmarshalAll/matches1000` 4011 allocs/~204740 B). Plausible tiny win on non-custom paths that previously paid the now-removed `field.Type().Implements()` call.

### Findings
- **[sand]** Correctness now rests on an unenforced invariant ("every value reaching `setFieldValue` is addressable"). All current callers honor it; a future caller passing a non-addressable value (map element, interface) would silently skip value-receiver dispatch. Optional: a one-line "callers must pass addressable values" note at the call sites would lock it in.
- **[dust]** `decoder.go:286` `|| fd.groupIndex == 0` dead condition (issue #109's 2nd item) is correctly deferred to #105 per the PR description.

### Tests
`go test ./...` green; `go vet` clean; bench suite compiles. `TestBenchmarkFixturesSane` passes all 7 subtests (no fixture rerouted). Value-/pointer-receiver dispatch benches + `TestUnmarshalRegexUnmarshaler/"value-receiver implementation is consulted"` all pass — proving value-receiver types still dispatch via the retained `CanAddr` branch.

<sub>Perf-focused review; branch rebased onto current main (clean) and force-pushed (`--force-with-lease`). 🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

Jecoms:
Addressed review feedback in a05ed61.

The two reviews disagreed on whether the deleted value-receiver dispatch branch was reachable. Settled it with an A/B test through the public `Unmarshal` API on an **interface-typed struct field** pre-populated with a concrete `RegexUnmarshaler`:
- main: passes (`UnmarshalRegex` runs)
- this branch before the fix: fails with `unsupported field type: interface`

So the branch was reachable — removing it regressed interface-typed fields (for interface-kind fields, `field.Addr().Interface()` is a `*interface` and doesn't satisfy `RegexUnmarshaler`, so the surviving `CanAddr` branch never fires). a05ed61 restores the `Type().Implements(...)` dispatch branch with a corrected comment and adds a regression test. Build/vet/test green.

This reframes the PR from "remove dead code" to "fix the misdiagnosis + add coverage." The earlier approving review's `CanAddr`/method-set reasoning holds only for concrete value types and didn't account for interface-kind fields.

Jecoms:
Reworked the PR framing to match the empirical finding:

- **Title** → `test(unmarshal): regression guard for interface-typed RegexUnmarshaler dispatch` (was `refactor: remove unreachable value-receiver dispatch branch`). The squash diff vs `main` is now just the regression test + clarifying comments — the removal and restoration cancel out, so `refactor` no longer described it (and Conventional-Commit CI would've validated the stale title).
- **Description** rewritten to explain that the `Implements` fallback is reachable for interface-typed fields, why (`field.Addr().Interface()` is a `*interface` that doesn't satisfy `RegexUnmarshaler`), and the A/B evidence.
- **`Closes #109` → `Refs #109`**: this PR reverses item 1 of that issue rather than resolving it, and item 2 (the `decoder.go` dead clause) is handled by #105.
- **Issue #109 corrected**: item 1 marked withdrawn (not dead) with the interface-field explanation; retitled/scoped to the remaining `|| fd.groupIndex == 0` decode-clause cleanup.

The approving review's `CanAddr`/method-set reasoning is sound but only for concrete field types — it didn't account for interface-kind fields. No code-logic change vs `main`; just the guard test and comments.

Jecoms:
Re-linked to **`Closes #109`** — both items of the issue are now accounted for:

- **Item 1** (`setFieldValue` fallback): resolved here via the regression test (the branch is reachable for interface-typed fields, not dead).
- **Item 2** (`decoder.go` `|| fd.groupIndex == 0`): already eliminated by #127's `groupIndexes`/occurrence-iteration rewrite — there's no surviving clause to remove, so no extra commit is needed in this PR.

No `decoder.go` change in this PR's net diff vs `main`, confirming item 2 is already done upstream. This PR fully addresses #109 on merge.

Issue #108, opened 2026-06-12
Refactor: route Unmarshal through a single shared decode plan (unify Unmarshal/Decoder semantics)


Jecoms:
Umbrella refactor addressing the root cause behind #104, #105, and #106: `Unmarshal`/`UnmarshalAll` and `Decoder` implement the same semantics twice and have drifted on edge cases (non-participating groups, duplicate names, case folding).

**Proposal:** have `Unmarshal`/`UnmarshalAll` build an uncached decoder plan and run the same `decode()` core the `Decoder` uses. One set of semantics, one place to fix bugs.

Side benefits this also picks up:
- **Per-match allocation waste in `UnmarshalAll`** (unmarshal.go:156-163): currently allocates a fresh `map[string]string` and re-walks `SubexpNames()` for every match, and `populateStruct` re-runs `parseFieldTag` + `strings.ToLower` per field per match. Hoisting a name->index plan once per call is a near-free win on the path `BenchmarkUnmarshalAll_logLines` measures.
- The group-map-building loop is copy-pasted three times (`Unmarshal`, `UnmarshalAll`, `NamedGroups`) — collapse to one helper.

Sequencing: land the targeted bug fixes first (so they're small and backportable), then this refactor to prevent re-divergence — or do the refactor first and get the bug fixes for free. Either way the end state is one decode core.

## Implementation plan

_Added 2026-06-30 via /gh-analyze-issue. Review before implementing._

**Status note.** The bug fixes (#104/#105/#106 + #123/#126/#127) already landed on `main`: both paths share `resolveGroupValue`, `strings.EqualFold` folding, and index-based "last participating occurrence wins". Side-benefit #2 is also done — the group-map loop is already collapsed into `namedGroupValues`/`fillNamedGroupValues` (regextra.go:402,414). What still diverges is the *mechanism*: Decoder runs a precomputed index plan (`[]fieldDecoder`); Unmarshal builds a per-match `map[string]string` and re-parses tags per field per match. This refactor unifies the mechanism only.

**Approach.** Extract the Decoder's plan/run into two non-generic free functions and route Unmarshal/UnmarshalAll through them. `compileDecoder[T]` is generic only for `reflect.TypeOf(zero)`; everything else is `reflect.Type`-driven, so the extraction is clean. A `strict bool` flag preserves the behavioral split: Decoder validates eagerly; Unmarshal stays lenient (skips undeclared-group / eager-default / layout-type errors).

**Steps.**
[...]

PR #127, opened 2026-06-12
fix: handle duplicate group names in map and decoder paths

### LINKED ISSUE #105: Bug: duplicate group names mishandled in both decode paths

Jecoms:
Fixes #105.

Patterns that reuse a group name (e.g. `(?:x(?P<word>a)|y(?P<word>b))`) were mishandled in opposite directions by the two paths:

- **NamedGroups / Unmarshal / UnmarshalAll** built their name→value map by iterating every group, so a non-participating later occurrence clobbered the real value with `""` — matching `"xa"` yielded `map[word:""]`.
- **Decoder** resolved each field via `SubexpIndex`, which reports only the *first* occurrence — decoding `"yb"` yielded a zero `Word`.

Changes:

- New shared `namedGroupValues` helper builds the map from `FindStringSubmatchIndex` pairs, which distinguish "did not participate" (negative indices) from "matched an empty span". A non-participating occurrence records `""` only if the name has no value yet. `NamedGroups`, `Unmarshal`, and `UnmarshalAll` all use it — this also collapses the map-building loop that was copy-pasted three times (a slice of the #108 refactor).
- `fieldDecoder` now precomputes `groupIndexes []int` (every occurrence, via new `subexpIndexes`) instead of a single `groupIndex`; `decode` scans them, last non-empty occurrence wins — same semantics as the map path. The dead `|| fd.groupIndex == 0` condition flagged in #109 disappears with the rewrite.

Semantics for *multiple participating* duplicates (sequential like `(?P<word>\w+) (?P<word>\w+)`) stay last-wins, now pinned by tests in both paths. `AllNamedGroups` is intentionally untouched — preserving every occurrence is its documented job, also pinned.

Tests: new `duplicate_groups_test.go` covering both alternation branches through `NamedGroups`, `Unmarshal`, `UnmarshalAll`, `Decoder.One`/`All`, sequential-duplicate last-wins parity, and `AllNamedGroups` preservation. Benchmarks still pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected decoding for regexes with the same named capture group repeated, so the resolved value is consistent across occurrences.
  * Fixed unmarshal behavior for optional typed groups: non-participating groups no longer cause errors and remain unset.
  * Improved named-group value handling for empty vs non-participating matches, including in named-group enumeration.
* **Performance**
  * Reduced allocation overhead during decoding and unmarshal by using index-based matching to distinguish participating vs non-participating groups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Duplicate group names no longer lose the participating value.** Go's `regexp` allows the same `(?P<name>...)` to appear more than once in a pattern (e.g. across alternation branches), where only one occurrence participates in a given match. Every name-based reader now resolves the occurrence that actually participated instead of trusting `re.SubexpIndex`'s first occurrence or letting a later empty/non-participating occurrence clobber a real value. Affects `NamedGroups`, `Unmarshal`, `UnmarshalAll`, `Decoder.One`/`Decoder.All`/`Decoder.Iter`, `FindNamed`, and `FindAllNamed`. For example, with `(?:x(?P<word>a)|y(?P<word>b))`, `FindNamed(re, "yb", "word")` now returns `("b", true)` (was `("", true)`) and `FindAllNamed(re, "xa yb", "word")` returns `["a", "b"]` (was `["a", ""]`). ([#127](https://github.com/Jecoms/regextra/pull/127), fixes [#105](https://github.com/Jecoms/regextra/issues/105))
+- **`Unmarshal`/`UnmarshalAll` no longer error on a non-participating optional group with a typed field.** A declared optional group that does not participate in the match (e.g. `(?P<num>\d+)?` against input with no digits) is now left at the field's zero value, matching `Decoder.One`, instead of feeding `""` into the numeric/bool/time conversion and returning `cannot convert "" to …`. Fields with a `default=` still substitute the default as before, and `NamedGroups` still reports a non-participating group as `""`. ([#127](https://github.com/Jecoms/regextra/pull/127))
+
+### Performance
+
+- Switching the `Decoder` to the index-returning match functions (`FindStringSubmatchIndex` / `FindAllStringSubmatchIndex`) — required to tell a participating-empty group from a non-participating one — also drops a per-match allocation: `Decoder.One` on a simple struct goes from 3 allocs to 2 (−40% B/op) and `Decoder.Iter` over 100 log lines from 309 allocs to 209 (−32%). ([#127](https://github.com/Jecoms/regextra/pull/127))
+
 ## [1.0.0] - 2026-05-05
 
 The API-stability stamp. Every public surface in `regextra.go`, `unmarshal.go`, and `decoder.go` was audited by the v1-readiness review (`docs/v1-readiness.md`) and carries a documented `Keep` verdict. The three "change before v1" blockers from that review (`AllNamedGroups` naming, tag-grammar reservation policy, no-match contract) all resolved as documentation lock-ins — every behavior in the v0.5.0 release ships unchanged into v1.0.0.
--- a/decoder.go
+++ b/decoder.go
@@ -45,9 +45,14 @@ type Decoder[T any] struct {
 type fieldDecoder struct {
 	// fieldIndex is the index into T's struct fields (StructField.Index[0]).
 	fieldIndex int
-	// groupIndex is the regex SubexpIndex for the named group; -1 means
-	// "no group declared, use default if present, otherwise skip."
-	groupIndex int
+	// groupIndexes holds the submatch index of every occurrence of the
+	// field's group name, in declaration order. Go's regexp allows the same
+	// group name to appear more than once in a pattern (e.g. across
+	// alternation branches), and only one occurrence participates in a given
+	// match — decode scans them all rather than trusting SubexpIndex's first
+	// occurrence. Empty means "no group declared, use default if present,
+	// otherwise skip."
+	groupIndexes []int
 	// opts is the parsed tag options map (e.g. {"default": "guest", "layout": "..."}).
 	// Nil if the field has no options.
 	opts map[string]string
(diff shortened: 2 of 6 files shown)


Jecoms:
## Code review

**Verdict:** The core fix is correct — non-participating duplicate occurrences no longer clobber participating values in `NamedGroups`/`Unmarshal`/`UnmarshalAll`, and the Decoder now scans all occurrences. But the two paths this PR claims to unify still disagree on one edge case, and the same bug class is left in two other public APIs.

### Findings

**1. [medium] Decoder and Unmarshal still diverge on participating-but-empty duplicates** — `decoder.go:306`
`Decoder.decode` uses "last *non-empty* occurrence wins" while `namedGroupValues` uses "last *participating* occurrence wins (even if empty)". The comment at `decoder.go:299-303` ("matching the Unmarshal path") is not accurate.
Repro: `(?P<word>a)(?P<word>b*)` on input `"a"` → `Decoder.One` gives `Word="a"`, but `Unmarshal`/`NamedGroups` give `""` (and with `default=dflt`, Unmarshal yields `"dflt"` while Decoder yields `"a"`). Root cause: `One`/`All`/`Iter` feed `decode` `FindStringSubmatch` strings, which can't distinguish participating-empty from non-participating, while Unmarshal now uses `FindStringSubmatchIndex`. Suggest switching the Decoder to the Index variants so both paths share one participation-aware semantic.

**2. [medium] `FindNamed`/`FindAllNamed` still have the issue-105 bug class** — `regextra.go:190`, `regextra.go:220`
Both still resolve duplicate names via `SubexpIndex`'s first occurrence. Repro: with `(?:x(?P<word>a)|y(?P<word>b))`, `FindNamed(re, "yb", "word")` returns `("", true)` even though `NamedGroups(re, "yb")["word"]` now correctly returns `"b"`; `FindAllNamed(re, "xa yb", "word")` returns `["a", ""]` instead of `["a", "b"]`. May be deliberately out of scope per the PR title, but nothing in the diff or docs records the remaining gap.

**3. [low] Pre-existing Unmarshal/Decoder divergence in this PR's territory** — `unmarshal.go:193`
`namedGroupValues` records `""` for a non-participating single optional group, and `populateStruct` passes that `""` to `setFieldValue` for typed fields: `y(?P<num>\d+)?` on `"y"` with `Num int` → Unmarshal errors `cannot convert "" to int`, while the equivalent `Decoder.One` succeeds with `Num=0`. Pre-existing on main, but this PR was the natural place to align or document it.

**4. [low] New doc cross-reference points to an API with the same ambiguity** — `regextra.go:241`
The new `NamedGroups` doc says "Use [AllNamedGroups] to see every occurrence", but `AllNamedGroups` still uses `FindStringSubmatch` and returns `""` for both non-participating and empty-matching occurrences — the exact conflation this PR fixed elsewhere.

**5. [low] No CHANGELOG entry** — `CHANGELOG.md`
This is a user-visible behavior fix (previously the last-declared occurrence always won, including `""`), but `[Unreleased]` is empty.

**6. [low] Dead defensive guard** — `decoder.go:307`
`gi < len(matches)` can never be false: `gi` comes from `SubexpNames` indices (max `NumSubexp`) and every `FindStringSubmatch` result has length `NumSubexp+1`. Drop it or comment it as an invariant assertion.

---

Jecoms:
Addressed in 0d1cd7e8.

**#1 (decoder/Unmarshal divergence) + #6 (dead guard) — fixed.** `Decoder.One/All/Iter` now match via `FindStringSubmatchIndex`, and `decode` applies the same "last *participating* occurrence wins, even if it matched an empty span" rule as `namedGroupValues`. The inaccurate comment is corrected and the `gi < len(matches)` guard is removed (the index slice length is invariant). `(?P<word>a)(?P<word>b*)` on `"a"` now yields `Word=""` from both paths — and `"dflt"` from both with `default=dflt`.

**#2 (FindNamed/FindAllNamed) — fixed.** Both now scan the group's occurrences and return the one that participated rather than `SubexpIndex`'s first: `FindNamed(re, "yb", "word")` → `("b", true)`, `FindAllNamed(re, "xa yb", "word")` → `["a", "b"]`. I also updated `TestFindNamed/found_second_group`, whose expectation (`"two"`) encoded the old first-occurrence behavior and contradicted `TestNamedGroups` (which expects the last, `"again"`).

**#3 (Unmarshal/Decoder typed-empty divergence) — aligned.** A non-participating optional group on a typed field is now left at its zero value by `Unmarshal`/`UnmarshalAll` (matching `Decoder.One`) instead of erroring on `"" → int`. Done via an `includeNonParticipating` flag on `namedGroupValues`, so `NamedGroups` keeps surfacing such groups as `""` while the unmarshal path omits them; `default=` still substitutes as before.

**#4 (AllNamedGroups cross-reference) — documented.** The `NamedGroups` doc now notes that `AllNamedGroups` reports a non-participating occurrence and an empty match identically (both `""`), so it can't tell them apart.

**#5 (CHANGELOG) — added.** `[Unreleased]` now carries Fixed + Performance entries.

**Verification.** `go vet`, full suite, `gofmt`, and `golangci-lint` (0 issues) all pass; regression tests added for each case above. Benchmarks *improve* rather than regress — `FindNamed` −50% B/op, `FindAllNamed` −28% B/op / −39% allocs, `Decoder.One` −40% B/op / −33% allocs, `Decoder.Iter` −32% allocs; time geomean flat (−0.7%, within run noise). All changes stay out of `setFieldValue` to avoid compounding the #122 merge conflict.

Issue #131, opened 2026-06-29
Release: prepare v2.0.0 — module /v2 path, changelog, breaking notes


Jecoms:
Cut the first major release since v1.0.0. Several merged changes alter observable behavior of existing calls — which README §Stability defines as breaking — so the next release is **v2.0.0** rather than a v1 patch/minor. This issue tracks the release prep so the tag lands on a commit that is correct for a Go v2+ module and a changelog that documents every change since v1.0.0.

**No migration guide.** There are no real v1 adopters, so we skip a migration write-up. The bar is an **exhaustive CHANGELOG** instead — every change since v1.0.0, with breaking ones clearly marked.

**Scope / known work**

- [x] Merge the remaining breaking PRs into the v2 line: #126 (case-fold unification), #123 (skip non-participating optional groups on struct reuse), and **#132** (`regex:"-"` exclude as a field, not name fallback — reverses the #85 "both mean no name" v1 promise, so it must ship on the v2 line, not a v1.x patch).
- [x] Go v2 module-path migration: `go.mod` `module …/regextra` → `…/regextra/v2`, the 8 external test imports, and README install/import/badge references. (Required Go v2 mechanics regardless of adoption; package never self-imports, so no non-test source rewrites.)
- [x] **Exhaustive CHANGELOG**: replace `[Unreleased]` with `## [2.0.0]` covering **every change since v1.0.0** — backfill the missing #122 and #129 entries, add #127, #125, #126, #123, #132, and audit `git log v1.0.0..main` so nothing is dropped. Mark breaking entries clearly (no separate migration guide needed).
- [x] README §Stability: bump "at v1" → v2 and refresh the breaking-change framing.
- [x] Release via the `release/v2.0.0` branch so `auto-tag.yml` publishes the tag + GitHub Release on merge.

**Must land before the cut (from open-issue triage)**

These are the only open issues that *must* ride the v2.0 tag rather than a later v2.x — because they are behavior-changing (locked-in once at the major boundary) or naturally bundle with release work already in flight:

- [x] **#119** — support `encoding.TextUnmarshaler` in field conversion. **Effectively breaking:** the full feature needs the `TextUnmarshaler` check *ahead* of the existing `time.Time`/`Kind` switches, which changes behavior for `time.Time` (RFC3339-only, ignores the `layout` tag) and named primitive types. That dispatch order is public contract and can't be tightened later without a v3 bump — so it must be settled at 2.0. (Medium risk; own PR + review.)
[...]

PR #132, opened 2026-06-29
fix(unmarshal)!: make regex:"-" exclude a field instead of name fallback

### LINKED ISSUE #106: Bug: `regex:"-"` does not exclude a field — it falls back to field-name matching
`parseFieldTag` (unmarshal.go:246-250) returns `""` for both an absent tag and `regex:"-"`, after which `findGroupValue` (unmarshal.go:283-294) and `matchGroupName` (decoder.go:111) fall back to matching by field name.

Everywhere else in Go (`encoding/json`, `encoding/xml`, `yaml`, `mapstructure`), `-` means "never populate this field." A user writing `regex:"-"` to opt a field out will still get it populated if a group happens to share the field's name. The current behavior is documented (regextra.go:134-135), but it's a trap that violates a universal convention.

**Fix:** make `"-"` mean "skip this field entirely," distinct from `""` (no tag -> name fallback).

**SemVer note:** this changes observable behavior of an existing call. Decide whether to classify it as a bug fix (aligning with documented Go-ecosystem convention) or hold for v2. Related: regextra.go:161 still says "regextra is pre-v1" while README/ROADMAP say v1 — see the stale-doc issue.

## Implementation plan

**SemVer decision: ship in the v2 line.** README §Stability (line 32) defines "changing observable behavior of an existing call" as breaking, and this fix does exactly that — a documented behavior (regextra.go:134-135) changes. So it cannot ship as a v1 patch/minor. It rides the v2.0.0 release being prepped under #131; land it on the v2 line and mark the CHANGELOG entry breaking.

**Approach.** Today `parseFieldTag` collapses `regex:"-"` and `regex:""` to the same `("", nil)` result, so both fall through to field-name matching. Give `parseFieldTag` a third return value, `skip bool`, set only for the bare-`-` tag. Both callers — `populateStruct` (Unmarshal) and `compileDecoder` (Decoder) — honor it: Unmarshal `continue`s the field; Decoder omits it from the decode plan. Only the whole-tag `"-"` skips; group names are Go identifiers so `-` can never name a real group, keeping the sentinel unambiguous.

**Steps.**
1. `parseFieldTag`: change signature to `(name string, opts map[string]string, skip bool)`; return `("", nil, true)` for `tag == "-"`, `("", nil, false)` for `tag == ""`.
2. `populateStruct` (unmarshal.go:193): capture `skip`; `if skip { continue }` before `findGroupValue`.
3. `compileDecoder` (decoder.go:111): capture `skip`; `if skip { continue }` before the name-fallback / group-resolution block, so the field never enters `d.fields`.
4. Update docs to say `-` excludes the field (distinct from `""`): package doc regextra.go:134-135, the `parseFieldTag` doc comment (unmarshal.go:231-232), and README tag-options section (~line 205-219).
5. Add a breaking CHANGELOG entry under the v2 section (coordinate with #131's exhaustive-changelog task).

**Files / surfaces touched.**
- `unmarshal.go` — `parseFieldTag` signature + `populateStruct` caller; doc comment.
- `decoder.go` — `compileDecoder` caller.
- `regextra.go` — package-doc tag-grammar wording (134-135).
- `README.md` — tag-options section: document `-` = exclude.
- `CHANGELOG.md` — breaking entry (v2).
- `*_test.go` — new exclusion tests (below).

**Verification.**
- `go test ./...` green.
- New tests for both paths: a field tagged `regex:"-"` whose name matches a declared group stays at its zero value (the trap case). Cover Unmarshal/UnmarshalAll and Decoder.One/All.
- Confirm `regex:""` (no-tag name fallback) still populates — guards against over-skipping.

**Open questions / risks.**
- Scope of the sentinel: recommend only the bare `"-"` tag skips. `regex:"-,default=..."` would parse `name="-"` (matches no group, default fires) — fine to leave as-is, but call it out in the doc so the bare-only rule is explicit. Flag if you'd rather treat any leading `-` as skip.
- The third return value touches every `parseFieldTag` call site — there are exactly two; the benchmarks reference it only in comments.

**References.**
- #131 — v2.0.0 release prep (where this breaking change lands).
- #112 — stale "pre-v1" package-doc claim referenced in the original report.
- README §Stability (breaking-change definition); regextra.go "Tag grammar" package doc.

Jecoms:
## Summary

- `regex:"-"` now **excludes** a field entirely, matching the `-` convention in `encoding/json`, `encoding/xml`, and `gopkg.in/yaml`. Previously `parseFieldTag` collapsed `regex:"-"` and `regex:""` to the same "no name" result, so a field tagged `regex:"-"` was still populated when a declared group happened to share its name — a trap that violated a universal Go convention.
- `parseFieldTag` gains a third return value (`skip bool`), set only for the bare `-` tag. Both callers — `populateStruct` (Unmarshal/UnmarshalAll) and `compileDecoder` (Decoder) — honor it and skip the field before any name fallback. An absent tag (`regex:""`) still falls back to the field name.
- Only the bare `-` excludes; `regex:"-,default=x"` still parses `-` as the group name (matches no group).
- Docs updated: package doc (`regextra.go`), `parseFieldTag` godoc, README tag-options section, and a breaking CHANGELOG entry.

**Breaking change** per README §Stability (it changes observable behavior of an existing call). Per the issue's plan and #131, this rides the v2.0.0 line; the PR title carries `!` and the CHANGELOG entry is marked breaking.

## Test plan

- [x] `go test -race ./...` — green
- [x] `go vet ./...`, `gofmt -s -l .`, `golangci-lint run` — clean (0 issues)
- [x] New `unmarshal_exclude_test.go` covers the trap case (field tagged `regex:"-"` whose name matches a group stays at zero value) across `Unmarshal`, `UnmarshalAll`, and `Decoder.One`/`All`, plus an over-skip guard confirming `regex:""` still falls back to the field name, and that a `regex:"-"` field needs no declared group at `Compile`.

Fixes #106

## Summary by CodeRabbit

* **Bug Fixes**
  * Fields tagged with `regex:"-"` are now fully excluded from decoding, so they won’t be populated even if a matching capture group exists.
  * Clarified tag behavior: `regex:""` falls back to the field name, while `regex:"-"` means “skip this field” unless combined with other options.
* **Documentation**
  * Updated the README and package docs to better explain the struct-tag syntax and the difference between empty and dash tags.


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`regex:"-"` now excludes a field instead of falling back to field-name matching. (Breaking.)** Previously `parseFieldTag` collapsed `regex:"-"` and `regex:""` to the same "no name" result, so a field tagged `regex:"-"` would still be populated if a declared group happened to share the field's name — the opposite of the `-` convention in `encoding/json`, `encoding/xml`, and `gopkg.in/yaml`. Now the bare `-` tag excludes the field entirely from both `Unmarshal`/`UnmarshalAll` and `Decoder`, leaving it at its zero value. An absent tag (`regex:""`) still falls back to the field's own name as before. Only the bare `-` excludes; a leading `-` followed by options (e.g. `regex:"-,default=x"`) still parses `-` as the group name. **Observable behavior change:** a field tagged `regex:"-"` whose name matches a group is no longer populated. ([#106](https://github.com/Jecoms/regextra/issues/106))
```
Jecoms: `[pebble]` This breaks behavior that was an explicitly locked-in v1 stability contract — commit c1325e1 (#85) documented `regex:""` and `regex:"-"` both as “no name” as a v1 promise. The `!` marker and “ride the v2.0.0 line” framing are right, but the v2 mechanism isn't in place yet: the `/v2` module-path migration tracked in #131 is unstarted, and #131's scope list (“merge remaining breaking PRs: #126, #123”) doesn't include this PR. Before the v2 tag is cut, fold #132 into #131's scope so it isn't released as a v1.x patch (which would be the actual semver break). (commit c1325e1 / #85; issue #131; README §Stability)

unmarshal.go
```diff
@@ -228,17 +232,27 @@ func populateStruct(structValue reflect.Value, groupValues map[string]string) er
...
-// `regex:""` and `regex:"-"` both signal "no name", returning "" and a nil
-// options map.
-func parseFieldTag(field reflect.StructField) (name string, opts map[string]string) {
+// The two forms differ:
+//   - `regex:""` (no tag) signals "no name", returning ("", nil, false); the
+//     caller falls back to matching the field's own name against a group.
+//   - `regex:"-"` signals "exclude this field", returning ("", nil, true); the
+//     caller skips the field entirely, never attempting a name fallback. This
+//     mirrors the `-` convention in encoding/json, encoding/xml, and
+//     gopkg.in/yaml. Only the bare `-` tag skips; a leading `-` followed by
```
Jecoms: `[dust]` This godoc says the caller “skips” the field (here and line 239), while the package doc, both inline `continue` comments, and the README all say “excludes.” Same action, two verbs in one block — worth unifying on “excludes” for consistency. (PR's own comments: regextra.go vs this godoc)

README.md
```diff
@@ -218,6 +218,8 @@ type LogLine struct {
 }
 ```
 
+**Excluding a field:** `regex:"-"` excludes a field entirely — it is never populated, even if a declared group happens to share the field's name. This matches the `-` convention in `encoding/json`, `encoding/xml`, and `gopkg.in/yaml`. It differs from an absent tag (`regex:""`), which falls back to matching the field's own name against a group. Only the bare `-` excludes; a leading `-` followed by options (e.g. `regex:"-,default=x"`) parses `-` as the group name, which matches no group.
```
Jecoms: `[dust]` Ends with “parses `-` as the group name, which matches no group.” The package doc and `parseFieldTag` godoc both add the reason (“since regexp group names are Go identifiers”). Mirroring that half-sentence here keeps the three doc locations aligned. (regextra.go package doc / `parseFieldTag` godoc)

Jecoms (commented):
### Code review

**6 finding(s)** — 1 pebble, 1 sand, 4 dust. Inline comments below.

#### PR-level

1. `[sand]` No test for the documented boundary that `regex:"-,default=x"` is *not* excluded — it parses `-` as the group name. The distinction is called out in three places (`parseFieldTag` godoc, README, CHANGELOG) but `unmarshal_exclude_test.go` only exercises the bare `-`. A one-line table case alongside `TestUnmarshal_emptyTagStillFallsBackToFieldName` would lock it in. (CONTRIBUTING.md “test edge cases” + the PR's own docs)

2. `[dust]` PR title uses `fix(unmarshal)!:`; the change also *adds* a capability that never existed (excluding a field via a tag), which fits `feat!:` under CONTRIBUTING.md's type table (`feat` = “new user-facing feature”). Defensible as `fix` since it resolves bug #106, and commitlint accepts both — surfaced only for awareness. (CONTRIBUTING.md commit-type table)

3. `[dust]` Exclusion is decided at compile time in `compileDecoder`, so `Decoder.Iter` inherits it mechanically, but prior reviews on this file (#123, #127) consistently asked for explicit `Iter` coverage when a compile/decode path changed. Optional. (prior-PR review pattern: #123, #127)

Jecoms:
Addressed in 4405390:

- **[dust] `parseFieldTag` godoc wording** — "skips" → "excludes" (both occurrences), so all five doc locations use one verb.
- **[dust] README reason** — exclude paragraph now ends "…matches no group since regexp group names are Go identifiers", matching the package doc and `parseFieldTag` godoc.
- **[sand] missing boundary test** — added `TestUnmarshal_dashWithOptionsIsNotExcluded`, asserting `regex:"-,default=unknown"` is *not* excluded (`-` is a group name that matches no group, so `default` applies).
- **[dust] Iter coverage** — added an `Iter` subtest to `TestDecoder_dashExcludesField`, per the #123/#127 convention.

Surfaced separately:
- **[pebble] v2 release scoping** — commented on #131 to add this PR to its "remaining breaking PRs" scope so the break ships on the v2 line, not a v1.x patch. No code change here.
- **[dust] `fix!` vs `feat!` title** — left as-is; flagged for awareness only and commitlint accepts both.

CHANGELOG.md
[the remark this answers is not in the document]
Jecoms: Good catch on the in-place case — fixed in 662475e, but with one correction.

The `Unmarshal` point is right: it populates the caller's struct in place (`populateStruct` just `continue`s on an excluded field) and never resets it, so a reused/pre-populated struct keeps the old value rather than going to zero.

But the suggested wording lumps `UnmarshalAll` in with `Unmarshal`, and that part isn't accurate: `UnmarshalAll` allocates a fresh slice via `reflect.MakeSlice` and replaces the whole thing (`elem.Set(newSlice)`), so its elements — and the excluded field — *are* the zero value. The real split is `Decoder` + `UnmarshalAll` (fresh result → zero) vs. `Unmarshal` (in place → unchanged), which is how the entry now reads.

PR #137, opened 2026-06-30
feat(unmarshal): support encoding.TextUnmarshaler in field conversion

### LINKED ISSUE #119: Feature: support encoding.TextUnmarshaler in field conversion
`setFieldValue` currently supports primitives, `time.Time`/`time.Duration`, and the package's own `RegexUnmarshaler`. Many common types already implement `encoding.TextUnmarshaler` — `netip.Addr`, `uuid.UUID`, `url.URL`, `big.Int`, `slog.Level` — and today users must write a `RegexUnmarshaler` wrapper for each.

**Proposal:** in `setFieldValue`, after the `RegexUnmarshaler` check (which should keep priority as the package-specific interface), check for `encoding.TextUnmarshaler` and call `UnmarshalText([]byte(value))`.

Explicitly non-breaking: the README's SemVer policy already classifies "accepting additional field types in Unmarshal/UnmarshalAll/Decoder" as a non-breaking addition. Mirror the existing dispatch-order documentation (unmarshal.go:349-352) for the new interface.

## Implementation plan

**Approach.** Add an `encoding.TextUnmarshaler` fallback to `setFieldValue` (unmarshal.go). It ranks *below* `RegexUnmarshaler` (package-specific extension point keeps priority) but, critically, must sit **after** the `time.Time`/`time.Duration` type switch, not immediately after the `RegexUnmarshaler` check: `time.Time` itself satisfies `encoding.TextUnmarshaler`, and its `UnmarshalText` only accepts RFC3339 — putting the new check earlier would silently drop the existing multi-layout `parseTime` fallback and the `layout` tag option. So the precedence becomes RegexUnmarshaler > time.Time/Duration special-case > TextUnmarshaler > built-in Kind switch. Dispatch mirrors the existing `RegexUnmarshaler` trio (addressable `field.Addr()`, plus `field.Type().Implements(...)` for interface-typed fields); pointer fields recurse into `Elem` and are caught by the addressable check.

**Steps.**
1. Add a cached `var textUnmarshalerType = reflect.TypeOf((*encoding.TextUnmarshaler)(nil)).Elem()` next to `regexUnmarshalerType`, and `import "encoding"`.
2. After the `time.Time`/`time.Duration` `switch field.Type()` block (~unmarshal.go:438) and before `switch field.Kind()`, add a numbered block: if `field.CanAddr()` and `field.Addr().Interface()` asserts to `encoding.TextUnmarshaler`, call `UnmarshalText([]byte(value))` and wrap any error (e.g. `cannot convert %q to %s: %w`); also handle interface-typed fields via `field.Type().Implements(textUnmarshalerType)`.
3. Update the dispatch-order comments in `setFieldValue` to document the new step, mirroring the numbered comments at unmarshal.go:382-411.
4. Docs: extend the "Supported field types" paragraph (README.md:169) and the `RegexUnmarshaler` section (README.md:230-240) to mention the `encoding.TextUnmarshaler` fallback and its precedence relative to `RegexUnmarshaler` and the time special-case.

**Files / surfaces touched.**
- `unmarshal.go` — new cached type var, the TextUnmarshaler dispatch block, updated dispatch-order doc comments.
- `README.md` — supported-types note + RegexUnmarshaler section precedence note (non-breaking per existing SemVer policy, no policy change needed).
- new `unmarshal_textunmarshaler_test.go` — table tests.

**Verification.**
- `go test ./...` with new cases: stdlib types that implement `TextUnmarshaler` and need *no* third-party dep — `netip.Addr`, `math/big.Int` (pointer receiver), `log/slog.Level` — as both value and pointer fields, plus failure cases (invalid input returns a wrapped error).
- Regression: confirm `time.Time` fields still use the multi-layout `parseTime` path and the `layout` opt (i.e. a non-RFC3339 layout like `2006-01-02` still parses), proving the ordering didn't regress.
- A type implementing *both* `RegexUnmarshaler` and `TextUnmarshaler` must still dispatch on `UnmarshalRegex`.

**Open questions / risks.**
- Ordering is the main risk — see Approach. Add the `time.Time` regression test to lock it in.
- The issue lists `url.URL` as a candidate, but `*url.URL` does **not** implement `encoding.TextUnmarshaler` (no `UnmarshalText`); don't use it as a test fixture. `uuid.UUID` is third-party — skip in tests to avoid a new dependency; mention it only in docs.
- Coverage gate is 92% (CI) — ensure the new branch (incl. error path and interface-typed field) is exercised.

**References.**
- Dispatch order + `RegexUnmarshaler`: unmarshal.go:362-483.
- SemVer non-breaking policy: README.md:38.

Jecoms:
## Summary
- Add an `encoding.TextUnmarshaler` fallback to `setFieldValue`, so any field whose type (or pointer-to-type) implements it — `netip.Addr`, `math/big.Int`, `log/slog.Level`, `uuid.UUID`, etc. — unmarshals with no `RegexUnmarshaler` wrapper.
- Precedence: `RegexUnmarshaler` > `time.Time`/`time.Duration` special-case > `TextUnmarshaler` > built-in Kind switch. The new check deliberately sits **after** the time special-case: `time.Time` satisfies `TextUnmarshaler` with an RFC3339-only `UnmarshalText`, so an earlier check would silently drop the multi-layout `parseTime` fallback and the `layout` tag option.
- Dispatch mirrors the existing `RegexUnmarshaler` trio (addressable `Addr()` + interface-typed `Implements`). Updated dispatch-order comments and README (supported-types note + `RegexUnmarshaler` precedence note). Non-breaking per the existing SemVer policy.

## Test plan
- [x] `go test -race ./...` — pass (new `unmarshal_textunmarshaler_test.go`: value & pointer fields, interface-typed field, error path, `RegexUnmarshaler`-beats-`TextUnmarshaler`, and a `time.Time` regression locking the ordering via a non-RFC3339 layout + fallback list)
- [x] `go vet ./...` — clean
- [x] `gofmt -s -l .` — clean
- [x] Coverage 96.8% (gate 92%); new branch incl. error path and interface-typed case exercised

Fixes #119

### DIFF
--- a/unmarshal.go
+++ b/unmarshal.go
@@ -1,6 +1,7 @@
 package regextra
 
 import (
+	"encoding"
 	"fmt"
 	"reflect"
 	"regexp"
@@ -359,6 +360,12 @@ type RegexUnmarshaler interface {
 // the implements-check on every field doesn't pay the reflect.TypeOf cost.
 var regexUnmarshalerType = reflect.TypeOf((*RegexUnmarshaler)(nil)).Elem()
 
+// textUnmarshalerType is the reflect.Type of encoding.TextUnmarshaler, cached
+// for the same reason as regexUnmarshalerType. Many stdlib and third-party
+// types (netip.Addr, big.Int, slog.Level, uuid.UUID, ...) already implement
+// it, so honoring it lets those drop into a struct field with no wrapper.
+var textUnmarshalerType = reflect.TypeOf((*encoding.TextUnmarshaler)(nil)).Elem()
+
 // setFieldValue sets the field value with appropriate type conversion.
 // `opts` carries per-field tag options parsed from `regex:"name,key=value,..."`.
 // Currently consulted: `layout` (for time.Time fields). Pass nil for no opts.
@@ -435,6 +442,32 @@ func setFieldValue(field reflect.Value, value string, opts map[string]string) er
 		}
 		field.Set(reflect.ValueOf(d))
 		return nil
+	}
+
+	// 3. encoding.TextUnmarshaler fallback. Ranks below RegexUnmarshaler (the
+	//    package-specific extension point keeps priority) AND below the
+	//    time.Time/time.Duration special-cases above: time.Time itself
+	//    satisfies TextUnmarshaler but its UnmarshalText only accepts RFC3339,
+	//    so checking it earlier would silently drop the multi-layout parseTime
+	//    fallback and the `layout` tag option. Mirrors the RegexUnmarshaler
+	//    dispatch: addressable fields via Addr(), interface-typed fields via
+	//    Implements. Pointer fields are handled in step 0 (recurse into Elem,
+	//    where the pointee is addressable and caught here).
+	if field.CanAddr() {
+		if u, ok := field.Addr().Interface().(encoding.TextUnmarshaler); ok {
+			if err := u.UnmarshalText([]byte(value)); err != nil {
+				return fmt.Errorf("cannot convert %q to %s: %w", value, field.Type(), err)
+			}
+			return nil
+		}
+	}
+	if field.Type().Implements(textUnmarshalerType) {
+		if u, ok := field.Interface().(encoding.TextUnmarshaler); ok {
+			if err := u.UnmarshalText([]byte(value)); err != nil {
+				return fmt.Errorf("cannot convert %q to %s: %w", value, field.Type(), err)
+			}
+			return nil
+		}
 	}
 
 	switch field.Kind() {
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ if err := regextra.Validate(re, "name", "age", "ssn"); err != nil {
 
 Unmarshal regex matches into a struct with automatic type conversion. Similar to `json.Unmarshal`, but for regex patterns.
 
-**Supported field types:** `string`, `int`, `int8`, `int16`, `int32`, `int64`, `uint`, `uint8`, `uint16`, `uint32`, `uint64`, `float32`, `float64`, `bool`, `time.Time`, `time.Duration`. Pointer-to-any-of-the-above is also supported — nil pointers are allocated, non-nil pointers are reused (pointee overwritten). For `time.Time`, several common layouts are tried (RFC3339, RFC3339Nano, `2006-01-02 15:04:05`, `2006-01-02`, `15:04:05`); `time.Duration` is parsed via `time.ParseDuration`. For caller-defined types, implement [`RegexUnmarshaler`](#regexunmarshaler-interface).
+**Supported field types:** `string`, `int`, `int8`, `int16`, `int32`, `int64`, `uint`, `uint8`, `uint16`, `uint32`, `uint64`, `float32`, `float64`, `bool`, `time.Time`, `time.Duration`. Pointer-to-any-of-the-above is also supported — nil pointers are allocated, non-nil pointers are reused (pointee overwritten). For `time.Time`, several common layouts are tried (RFC3339, RFC3339Nano, `2006-01-02 15:04:05`, `2006-01-02`, `15:04:05`); `time.Duration` is parsed via `time.ParseDuration`. Any field whose type (or pointer-to-type) implements [`encoding.TextUnmarshaler`](https://pkg.go.dev/encoding#TextUnmarshaler) is also supported out of the box — e.g. `netip.Addr`, `math/big.Int`, `log/slog.Level`, `github.com/google/uuid.UUID` — by calling its `UnmarshalText` with the matched value. For caller-defined types, implement [`RegexUnmarshaler`](#regexunmarshaler-interface).
 
 **Field mapping priority:**
 1. Struct tag `regex:"groupname"` if provided (highest priority)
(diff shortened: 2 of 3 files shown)


unmarshal.go
```diff
@@ -437,6 +444,32 @@ func setFieldValue(field reflect.Value, value string, opts map[string]string) er
...
+		}
+	}
+	if field.Type().Implements(textUnmarshalerType) {
+		if u, ok := field.Interface().(encoding.TextUnmarshaler); ok {
+			if err := u.UnmarshalText([]byte(value)); err != nil {
+				return fmt.Errorf("cannot convert %q to %s: %w", value, field.Type(), err)
+			}
+			return nil
+		}
+	}
```
Jecoms, line 471: `[dust]` The two `TextUnmarshaler` branches duplicate the same `UnmarshalText` + error-wrap body, and the whole block structurally mirrors the `RegexUnmarshaler` dispatch above. A tiny local helper could DRY both pairs, but this intentionally matches the existing `RegexUnmarshaler` style, so it's optional and arguably more readable left as-is. (consistency with existing dispatch in unmarshal.go; CONTRIBUTING.md "Prefer simplicity over cleverness")

unmarshal.go
```diff
@@ -437,6 +444,32 @@ func setFieldValue(field reflect.Value, value string, opts map[string]string) er
...
+	//    where the pointee is addressable and caught here).
+	if field.CanAddr() {
+		if u, ok := field.Addr().Interface().(encoding.TextUnmarshaler); ok {
+			if err := u.UnmarshalText([]byte(value)); err != nil {
+				return fmt.Errorf("cannot convert %q to %s: %w", value, field.Type(), err)
+			}
+			return nil
+		}
+	}
+	if field.Type().Implements(textUnmarshalerType) {
```
Jecoms, line 464: `[dust]` Awareness only: a field whose type has a numeric/string underlying Kind *and* implements `TextUnmarshaler` (e.g. `slog.Level`) is now intercepted here before the Kind switch, so a numeric-string input that previously parsed via `SetInt` would route through `UnmarshalText` instead. This is intentional and matches `encoding/json` precedence — just noting the subtle behavior shift for such types.

Jecoms (commented):
### Code review

**3 finding(s)** — 1 sand, 2 dust. Inline comments below.

Solid, well-scoped PR. The core design constraint holds: the new `encoding.TextUnmarshaler` dispatch (step 3) sits *below* the `time.Time`/`time.Duration` special-case (step 2), so the multi-layout `parseTime` fallback and the `layout` tag option are preserved — `TestUnmarshal_TimeBeatsTextUnmarshaler` locks exactly this ordering. The dual-branch dispatch (`CanAddr().Addr()` for concrete fields + `Type().Implements` for interface-typed fields) correctly mirrors the `RegexUnmarshaler` pattern hardened in #125, and the matching `InterfaceField` test guards the branch #125 proved is load-bearing. Pointer fields route through step 0 and recurse into the addressable pointee, which the value/pointer tests confirm. `go test -race`, `go vet`, and `gofmt -s` all clean locally.

#### PR-level

1. `[sand]` No godoc `Example` was added for the new capability. The repo consistently ships a focused example per Unmarshal feature (`ExampleRegexUnmarshaler`, `ExampleUnmarshal_timeTypes`, `ExampleUnmarshal_pointerFields`), and CONTRIBUTING calls for example tests that demonstrate usage / appear in godoc. An `ExampleUnmarshal_textUnmarshaler` (e.g. `netip.Addr`) would match the established pattern and surface the feature on pkg.go.dev. (CONTRIBUTING.md "Documentation" + "Testing")

Jecoms:
Addressed review in 3ed73d6:

- `[sand]` Added `ExampleUnmarshal_textUnmarshaler` (godoc/runnable, using stdlib `netip.Addr`) so the new TextUnmarshaler capability follows the per-feature example convention and surfaces on pkg.go.dev. Placed alongside the other #119 tests in `unmarshal_textunmarshaler_test.go`.

Left as-is (intentional, as the review noted):

- `[dust]` Duplicate `UnmarshalText` + error-wrap body across the two dispatch branches — kept to mirror the existing `RegexUnmarshaler` dispatch style (consistency / "prefer simplicity over cleverness").
- `[dust]` Numeric/string-underlying types implementing `TextUnmarshaler` (e.g. `slog.Level`) now intercepted before the Kind switch — intentional, matches `encoding/json` precedence; already locked by `TestUnmarshal_TextUnmarshaler_Value`.

Local checks: `go test -race ./...` pass, `go vet ./...` clean, `gofmt -s -l .` clean, `golangci-lint run` 0 issues.

unmarshal.go
Jecoms: Intentional — keeping the duplicated dispatch body mirrors the existing `RegexUnmarshaler` block right above it, which keeps the precedence/ordering easy to read in one place and matches CONTRIBUTING's "prefer simplicity over cleverness". Leaving as-is rather than introducing a helper. Thanks for the note.

unmarshal.go
Jecoms: Confirmed intentional — the `TextUnmarshaler` interception before the Kind switch is the desired precedence (matches `encoding/json`), and it's locked by `TestUnmarshal_TextUnmarshaler_Value` so the behavior shift for types like `slog.Level` won't regress silently. Documented the precedence in the README too. Thanks for flagging the subtlety.

PR #140, opened 2026-06-30
feat: add ReplaceFunc for callback-based named-group replacement

### LINKED ISSUE #117: Feature: ReplaceFunc — callback-based replacement for named groups
`Replace` only accepts a static `map[string]string`, so the replacement can't depend on the matched value. Redaction and normalization use cases need the matched text:

```go
// proposed
func ReplaceFunc(re *regexp.Regexp, s string, fn func(group, match string) string) string
```

e.g. masking all but the last 4 digits of a captured card number, or lowercasing captured hostnames. This is the highest-value functionality gap found in review — the stdlib has `ReplaceAllStringFunc` but it operates on whole matches, not named groups.

Additive, non-breaking. Should share the span-collection logic with `Replace` (and inherit the stable-sort fix).

## Implementation plan

**Approach.** Extract the span collect → stable-sort (start asc, end desc) → outermost-wins apply loop from `Replace` into one unexported helper, then make both `Replace` and `ReplaceFunc` thin wrappers over it. Store `{start, end, name}` per span and resolve the replacement string *at apply time* (only for spans that actually win the overlap), so `fn` is never invoked for inner groups skipped by the outermost-wins rule. `Replace`'s map lookup moves to the same apply point — behavior-equivalent, and both share the #107 fix for free.

**Steps.**
1. Add helper `replaceNamed(re, target, replFor func(name, matched string) (string, bool)) string` holding the existing loop + `slices.SortStableFunc`; `ok=false` skips a span (group passes through).
2. Refactor `Replace` to call it with a map-lookup closure; verify existing tests/bench unchanged.
3. Add `ReplaceFunc` calling it with `func(name, m string) (string, bool) { return fn(name, m), true }` — `fn` runs once per applied named span, left-to-right.
4. Godoc + README + package-doc "API at a glance" / "No-match behavior" table entry.

**Files / surfaces touched.**
- `regextra.go` — new helper, refactored `Replace`, new `ReplaceFunc`, package-doc updates.
- `regextra_test.go` — `TestReplaceFunc` (table-driven) + `ExampleReplaceFunc`.
- `regextra_bench_test.go` — `BenchmarkReplaceFunc`.
- `README.md` — `ReplaceFunc` section after `Replace`.

**Verification.**
- `go test -race ./...`: redaction/normalization cases, no-match (returns target, `fn` never called), overlap/nesting (inner-group `fn` not invoked), duplicate/non-participating group names. Confirm `Replace` benchmark/allocs don't regress after the refactor.

**Open questions / risks.**
- Contract for "leave unchanged": caller returns `match` verbatim — document explicitly (unlike `Replace`, every participating named group reaches `fn`).
- `nil fn`: document as programmer error (panic), mirroring stdlib `ReplaceAllStringFunc`.
- Resolving the closure at apply time adds one indirection per applied span; acceptable, but watch the `Replace` bench.

**References.**
- `Replace` impl + #107 stable-sort fix in `regextra.go`; stdlib `Regexp.ReplaceAllStringFunc` (whole-match analogue).
- ROADMAP "Next minor" additive nice-to-haves (sibling to `ReplaceFirst`).

Jecoms:
## Summary
- Add `ReplaceFunc(re, target, fn func(group, match string) string)` — a callback analogue of `Replace` whose substitution depends on the matched value (redaction, normalization), filling the gap left by `Replace`'s static `map[string]string`.
- Extract the span collect → stable-sort (start asc, end desc) → outermost-wins apply loop into a shared unexported `replaceNamed` helper; `Replace` and `ReplaceFunc` are now thin wrappers and both inherit the #107 stable-sort tie-break.
- Resolve the replacement at apply time, so `fn` is never invoked for an inner group suppressed by an enclosing outermost span. `Replace`'s map lookup moves to the same point (behavior-equivalent).
- Docs: godoc on `ReplaceFunc`/`replaceNamed`, package-doc "API at a glance" + "No-match behavior" table entries, and a README section after `Replace`.

Contract notes (per the issue's implementation plan): every participating named group reaches `fn`; return the match verbatim to leave a group unchanged. On no match the target is returned unchanged and `fn` is never called. A nil `fn` is a programmer error and panics on the first match, mirroring stdlib `Regexp.ReplaceAllStringFunc`.

## Test plan
- [x] `go test -race ./...` — all pass (table-driven `TestReplaceFunc` covering redaction, normalization, group-name passthrough, no-match, non-matching-text preservation, nested/overlap, optional non-participating, duplicate names; plus targeted tests that `fn` is not invoked for a suppressed inner group and never called on no match; `ExampleReplaceFunc`).
- [x] `go vet ./...`, `gofmt -s -l .`, `golangci-lint run` — clean.
- [x] `go test -bench BenchmarkReplace -benchmem` — added `BenchmarkReplaceFunc`; existing `Replace` benchmarks remain in line after the refactor.

Fixes #117

### DIFF
--- a/regextra.go
+++ b/regextra.go
@@ -43,6 +43,7 @@ By use case:
   - Pull every named group from one match, keeping every value when a group
     name is reused inside the pattern (map of slices): [AllNamedGroups]
   - Substitute named-group spans by name: [Replace]
+  - Substitute named-group spans with a callback over the matched value: [ReplaceFunc]
   - Assert at startup that required groups are declared: [Validate]
   - Decode one match into a struct: [Unmarshal]
   - Decode all matches into a slice of structs: [UnmarshalAll]
@@ -73,6 +74,8 @@ the no-match form that lets the caller continue without a special-case branch.
 	                                          name is not declared on the regex)
 	NamedGroups, AllNamedGroups               empty map (initialized, not nil)
 	Replace                                   target returned unchanged
+	ReplaceFunc                               target returned unchanged (fn
+	                                          never called)
 	Validate                                  unrelated — checks declarations,
 	                                          not matches against a target
 	Unmarshal                                 nil error; destination struct left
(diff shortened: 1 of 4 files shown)


Jecoms (commented):
### Code review

**2 finding(s)** — 1 pebble, 1 dust. Inline comments below.

regextra.go
```diff
@@ -424,6 +427,57 @@ func Replace(re *regexp.Regexp, target string, replacements map[string]string) s
...
+// looked up in a static map. Use it for substitutions that depend on what
+// matched: redaction (mask all but the last four digits of a card number),
+// normalization (lowercase a captured host), and similar.
+//
+// fn runs once per substituted named span, left to right. To leave a group
+// unchanged, return its match verbatim. This differs from [Replace], which
+// passes a group through when it is absent from the map: every participating
+// named group reaches fn, so fn itself decides what to keep. Passing a nil fn
+// is a programmer error — it panics on the first match, mirroring the standard
+// library's [regexp.Regexp.ReplaceAllStringFunc].
```
Jecoms, line 449: `[pebble]` The nil-`fn` contract documented here (panics on the first match, but never on no match) is untested. CONTRIBUTING explicitly lists nil pointers and "no matches" as edge cases to cover and asks for critical error paths to be tested; the no-match-with-nil-`fn` path (returns target unchanged, never panics) is a precise, easily-pinned guarantee worth a test. (CONTRIBUTING.md — "Test edge cases (nil pointers, empty strings, no matches)" / "Critical error paths must be covered")

regextra.go
```diff
@@ -424,6 +427,57 @@ func Replace(re *regexp.Regexp, target string, replacements map[string]string) s
...
+// and the text it matched. Like [Replace] it operates on every match of re, in
+// order — but the replacement is computed from the matched value rather than
+// looked up in a static map. Use it for substitutions that depend on what
+// matched: redaction (mask all but the last four digits of a card number),
+// normalization (lowercase a captured host), and similar.
+//
+// fn runs once per substituted named span, left to right. To leave a group
+// unchanged, return its match verbatim. This differs from [Replace], which
+// passes a group through when it is absent from the map: every participating
+// named group reaches fn, so fn itself decides what to keep. Passing a nil fn
```
Jecoms, line 447: `[dust]` "every participating named group reaches fn" is slightly imprecise: a participating inner group inside an already-substituted outermost span does not reach `fn` (as the overlap paragraph two lines below states). Reads as a minor internal tension; consider "every participating named group not suppressed by an outermost span". This is the wording agreed in the issue plan, so low priority. (godoc on ReplaceFunc, regextra.go)

regextra.go
Jecoms: Addressed in 7fe726b — added `TestReplaceFuncNilFn` to regextra_test.go with two subtests: "panics on first match" (nil fn with a match recovers a non-nil panic) and "no panic on no match" (nil fn on "no matches here" returns target unchanged and never panics). `go test ./...` and `go test -race` pass.

regextra.go
Jecoms: Leaving as-is — this is the wording agreed in the issue plan, and the apparent tension is resolved by the overlap paragraph two lines below ("fn is not called for inner groups inside an already-substituted span") plus the replaceNamed doc. Keeping the agreed phrasing to avoid contradicting the issue; happy to tighten in a follow-up if you'd prefer the "not suppressed by an outermost span" qualifier.

PR #142, opened 2026-06-30
feat: add DecodeError type and standardize error prefixes (#111)

### LINKED ISSUE #111: Cleanup: standardize error prefixes; add a typed error for conversion failures
Error formatting is inconsistent across the three paths:

- `Decoder.One` surfaces bare `field Age: ...` (decoder.go:300) with no `regextra` prefix
- `Decoder.All` wraps with `regextra.Decoder.All: match %d: ...` (decoder.go:228)
- The Unmarshal path uses `regextra: failed to set field X` (unmarshal.go:217)

Pick one prefix convention and apply it everywhere.

Additionally, conversion failures have no typed error — only `ErrNoMatch` is programmatically distinguishable. Consider a `*DecodeError` (carrying field name, group name, raw value, wrapped cause) so callers can branch without string matching. The README explicitly tells users not to pattern-match `err.Error()`, which only works if sentinels/types exist for the cases callers care about. ROADMAP already plans typed `*ValidationError` for Compile — design these together.

## Implementation plan

**Approach.** Adopt one prefix convention for all constructed errors — the method-qualified `regextra.<Entrypoint>:` form already used by `Compile`, `Validate`, and `Decoder.All` — and apply it to the two stragglers (`Unmarshal`/`UnmarshalAll`'s bare `regextra:` and `Decoder.One`'s prefix-less `field %s:`). Separately, introduce an exported `*DecodeError` value type carrying the field name, group name, raw value, target type, and wrapped cause, constructed at the single shared decode site so both the reflect (`Unmarshal`) and typed (`Decoder`) paths return the same `errors.As`-able error. Entrypoints then wrap the `*DecodeError` with their `regextra.<Entrypoint>:` prefix; `%w` keeps the typed error reachable. Shape `*DecodeError` so the roadmap's planned `*ValidationError` (for `Compile`) can mirror it.

**Steps.**
1. Decide and document the convention: `regextra.<ExportedEntrypoint>:` prefix for contextual/constructed errors; existing package sentinels (`ErrNoMatch` = `regextra: no match`) keep the bare `regextra:` prefix since a sentinel value isn't bound to one entrypoint. (Renaming message text is explicitly non-breaking — README "Stability".)
2. Define `DecodeError` in `unmarshal.go` (the conversion surface): exported struct with `Field`, `Group`, `Value`, `Type string`, `Err error`; `Error()` and `Unwrap()` methods. Keep the low-level `setFieldValue` returning the raw "cannot convert…" cause; wrap into `*DecodeError` at the two callers that know field/group/value.
3. In `populateStruct` (unmarshal.go), replace `regextra: failed to set field %s: %w` with a `*DecodeError`; have `Unmarshal`/`UnmarshalAll` apply the `regextra.Unmarshal:` / `regextra.UnmarshalAll: match %d:` prefix (also restandardize their bare-`regextra:` argument-validation errors).
4. In `decode` (decoder.go), build a `*DecodeError` instead of `fmt.Errorf("field %s: %w")`; add a `groupName string` to `fieldDecoder` (currently only `groupIndexes` is retained after Compile) so the decoder path can populate `Group`. Have `One`/`Iter` add the `regextra.Decoder.One:` / `regextra.Decoder.Iter:` prefix; `All` already prefixes correctly.
5. Update README error/stability sections (lines ~39, ~370) to document `errors.As(err, &de)` for `*DecodeError`; add a CHANGELOG entry. Optionally note the `*DecodeError`/`*ValidationError` symmetry near the roadmap item.

**Files / surfaces touched.**
- `unmarshal.go` — new `DecodeError` type; restandardize `Unmarshal`/`UnmarshalAll` prefixes (lines ~88,93,133,138,144,209); wrap conversion failures as `*DecodeError`.
- `decoder.go` — add `groupName` to `fieldDecoder`; `decode` returns `*DecodeError`; `One`/`Iter` prefix wrapping (line ~324); `Compile`/`All` already conform.
- `unmarshal_test.go`, `decoder_test.go` — assert `errors.As(err, &de)` and populated fields via the public API on both paths; assert prefix presence loosely (don't pin full wording, per Stability).
- `README.md`, `CHANGELOG.md` — document the typed error and convention.

**Verification.**
- `go test -v -race ./...`, `golangci-lint run`, `gofmt -s -l .`, `go vet ./...` all clean; coverage stays >92%.
- New tests prove a conversion failure surfaces a `*DecodeError` with correct `Field`/`Group`/`Value`/`Type` from `Unmarshal`, `UnmarshalAll`, `Decoder.One`, `Decoder.All`, and `Decoder.Iter`, and that every entrypoint's error string carries its `regextra.<Entrypoint>:` prefix.

**Open questions / risks.**
- Convention exception: confirm sentinels (`ErrNoMatch`) keep bare `regextra:` while contextual errors use `regextra.<Entrypoint>:`. The alternative (bare `regextra:` everywhere) is more churn and drops the useful method context already present on `Compile`/`Validate`/`Decoder.All`.
- `Type` field representation: `string` (rendered) vs `reflect.Type` — `string` keeps the public struct dependency-light and avoids exposing reflect in the API; prefer `string`.
- Adding an exported type is additive/non-breaking, but `*DecodeError` becomes v1 API surface — its field set must be right the first time and should template `*ValidationError`.

**References.**
- ROADMAP.md "Next minor" — typed `*ValidationError` for `Compile`; design `*DecodeError` to mirror it.
- README.md "Stability" (line 39) — message-wording changes are non-breaking; callers compare sentinels/types, not `err.Error()`.
- Existing sentinel precedent: `ErrNoMatch` (decoder.go:15).

Jecoms:
## Summary
- Standardize constructed/contextual error prefixes on `regextra.<Entrypoint>:` — fixes the stragglers: `Unmarshal`/`UnmarshalAll`'s bare `regextra:` argument-validation errors and `Decoder.One`/`Iter`'s previously-prefixless conversion errors. Package sentinels (`ErrNoMatch`) keep the bare `regextra:` prefix.
- Add exported typed error `*DecodeError{Field, Group, Value, Type, Err}` with `Error()`/`Unwrap()`, constructed at the two shared decode sites (`populateStruct` in unmarshal.go, `decode` in decoder.go) so both the reflect (`Unmarshal`) and typed (`Decoder`) paths return the same `errors.As`-able error. Entrypoints wrap it with their `regextra.<Entrypoint>:` prefix via `%w`.
- `fieldDecoder` now retains the resolved `groupName` (previously only `groupIndexes` survived Compile) so the decoder path can populate `DecodeError.Group`; `findGroupValue` reports the resolved group name for the reflect path.
- Docs: README Stability + Decoder error sections document `errors.As(err, &de)`; CHANGELOG Added/Changed entries.

Shaped `*DecodeError` to template the roadmap's planned `*ValidationError` for `Compile`. Adding an exported type is additive/non-breaking; message-wording changes are non-breaking per README §Stability.

## Test plan
- [x] `go test -race ./...` — pass
- [x] New tests assert `errors.As(err, &de)` with populated `Field`/`Group`/`Value`/`Type` and the entrypoint prefix on `Unmarshal`, `UnmarshalAll`, `Decoder.One`, `Decoder.All`, and `Decoder.Iter` (prefix asserted loosely per Stability)
- [x] `gofmt -s -l .` clean, `go vet ./...` clean, `golangci-lint run` — 0 issues
- [x] Coverage 97.6% (>92% gate)

Fixes #111

## Summary by CodeRabbit

* **New Features**
  * Added an exported typed `DecodeError` for capture-group conversion failures (field, group, raw value, target type, wrapped cause).
  * Named-group extraction now returns named-group maps for every match (aligned with existing `NamedGroups` semantics).

* **Bug Fixes**
  * Unmarshal/UnmarshalAll now use the same shared decode plan as the decoder, unifying skip/default and field-mapping behavior.
  * Improved error wrapping and standardized `regextra.<entrypoint>:` prefixes, including match-specific context.

* **Documentation**
  * Updated SemVer guidance and added “Typed conversion failures” documentation with `errors.As` examples.


unmarshal.go
```diff
@@ -46,6 +46,48 @@ func parseTime(value string) (time.Time, error) {
...
+// Err holds the underlying conversion cause (e.g. a *strconv.NumError or a
+// time-parsing error) and is reachable via [errors.Is]/[errors.As] through
+// Unwrap. No match is not a DecodeError — [Unmarshal]/[UnmarshalAll] return nil
+// and [Decoder.One] returns [ErrNoMatch] in that case.
+type DecodeError struct {
+	// Field is the destination struct field name.
+	Field string
+	// Group is the capture group the value was read from. It is the field's
+	// `regex:"..."` tag name when set, otherwise the group resolved from the
+	// field name. Empty only when a field with a `default=` maps to no group.
```
Jecoms: `[sand]` This "Empty only when a field with a `default=` maps to no group" claim doesn't hold on either path. On the reflect path, `findGroupValue` falls back to the attempted key, so a default-only no-group field reports `Group=<field name>` (verified: `Group="Count"`), never empty. On the typed path, defaults are pre-validated in `compileDecoder`, so a default-only field can't produce a runtime `*DecodeError` at all — the empty case is effectively unobservable, and the two paths would populate `Group` differently if it ever were. Consider tightening the doc (and reconciling the two paths' `Group` rules) since this is new v1 API surface. (issue #111: "same `errors.As`-able error" goal; verified against both code paths)

Jecoms (commented):
### Code review

**2 finding(s)** — 1 sand, 1 dust. Inline comments below.

Verified locally on `95650f3`: `go build`, `go vet`, `gofmt -s -l .`, and `golangci-lint run` (0 issues) all clean; `go test -race -covermode=atomic ./...` passes at 97.6% coverage. The typed-error wiring is correct on both paths — `populateStruct` (reflect) and `decode` (typed) construct the same `*DecodeError` with `Field`/`Group`/`Value`/`Type` populated and the cause preserved via `%w`/`Unwrap`; entrypoints wrap with their `regextra.<Entrypoint>:` prefix; `ErrNoMatch` correctly keeps the bare `regextra:` prefix, and a grep confirms it is the only remaining bare-prefix constructed error. New tests exercise all five entrypoints (`Unmarshal`, `UnmarshalAll`, `Decoder.One`/`All`/`Iter`) plus the arg-validation prefixes.

#### PR-level

1. `[dust]` Back-compat awareness: the prior `regextra: failed to set field %s` wording and the bare `regextra:` argument-validation strings changed (now `regextra.Unmarshal: field %s: …` / `regextra.UnmarshalAll: match N: …`). This is sanctioned non-breaking per README §Stability (compare sentinels/types, not strings), but any downstream `strings.Contains` on the old text would break — the new `*DecodeError` + `errors.As` is the intended migration path. (README.md §Stability)

unmarshal.go
Jecoms: Addressed in 0262ece. Reworded the `DecodeError.Group` godoc to drop the inaccurate "empty only when a `default=` field maps to no group" claim. Confirmed against the code:

- Reflect path (`populateStruct` → `findGroupValue`): the no-match branch returns the field name as the resolved key, so `Group` is never empty (e.g. `Group="Count"`).
- Typed path (`decode` → `fd.groupName`): a default-only field that maps to no group does carry `groupName=""`, but `compileDecoder` pre-validates the `default=` (probe `setFieldValue`), so the same value can't fail at runtime — that field never produces a `*DecodeError`, making the empty case unobservable.

New doc states Group is populated on every observable DecodeError and spells out how each path reaches the unset case. `go build`/`go test -race`/`gofmt -s -l`/`go vet`/`golangci-lint run` all clean.

Jecoms:
Addressed the review (0262ece):

- **`[sand]` `DecodeError.Group` godoc** — reworded to accurately describe when `Group` is set vs. empty per path. See the inline reply for the verification.
- **`[dust]` back-compat awareness on changed error strings** — acknowledged, no code change. The wording change is sanctioned non-breaking per README §Stability (compare sentinels/types, not strings; the new `*DecodeError` + `errors.As` is the migration path). This is already documented in CHANGELOG.md (Unreleased), which calls out that `Unmarshal`/`UnmarshalAll` argument-validation errors moved off the bare `regextra:` prefix and `Decoder.One`/`Iter` gained the `regextra.<Entrypoint>:` prefix, with the non-breaking-wording note. No CHANGELOG addition needed.

Verification on the new commit: `go build ./...`, `go test -race ./...`, `gofmt -s -l .`, `go vet ./...`, `golangci-lint run` (0 issues) — all clean.

Jecoms:
## Benchmark A/B: before vs after #111 (DecodeError + standardized prefixes)

**Method.** Full suite on the PR parent `7e36ad2` (base) vs PR head `0262ece` (after), compared with `benchstat`. Base ran the whole suite (110 cases); head ran the benchmarks that exercise the changed code paths (`Unmarshal`/`Compile`/`Decoder`) plus `Validate` as an untouched control. Runs were **serial** (no CPU contention). `-benchmem -count=10`, `go test -timeout=30m`.

```
goos: darwin   goarch: arm64   cpu: Apple M4   go: 1.24.0
```

**Noise floor (untouched controls):** `Compile/regexpCompileBaseline` +1.57% time / 0 B, `Validate` −3.1%…+0.8% / 0 alloc, `DecoderPattern` −3.2% (≈1ns). So **time deltas under ~3% are noise**; alloc/byte deltas are exact (`±0%`).

### TL;DR
- **Hot success path (decode / unmarshal of matching input): unchanged** — identical `allocs/op` and `B/op`, time within noise.
- **`Compile` (one-time per pattern): small real regression** — same alloc count, but wider allocations and ~+7% time, from the new `groupName string` on every `fieldDecoder`.
- **Conversion-error path: real, intended** — `+2 allocs/op` per *failed* conversion, the cost of building the new `*DecodeError`.
- **Net:** no hot-path regression; no consumer concern. The added cost lands only on `Compile` (amortized away by compile-once-reuse) and on the failure path.

### 1. Hot success path — not affected
| benchmark | sec/op | B/op | allocs/op |
|---|---|---|---|
| `Unmarshal/simpleThreeField` | +0.49% (noise) | −0.21% | ~ (p=1.00) |
| `Unmarshal/wideStruct` | +1.36% (noise) | ~ | ~ |
| `DecoderOne/manyFields` | +1.52% (noise) | ~ | ~ |
| `DecoderAll/matches1000` | +2.92% (noise) | ~ | ~ |
| `UnmarshalAll/matches100` | −1.56% | −0.20% | ~ |

`allocs/op` and `B/op` are identical across the whole success path; the few % on time sits in the ±3% noise floor.

### 2. `Compile` — one-time regression (same allocs, bigger + slower)
| benchmark | sec/op | B/op | allocs/op |
|---|---|---|---|
| `CompilePlanOnly` | +7.84% | **+25.45%** (440→552 B) | ~ (unchanged) |
| `Compile/simpleStruct` | +7.44% | +2.13% | ~ |
| `Compile/wideStruct` | +7.04% | +4.26% | ~ |
| `Compile/fields20` | +3.92% | +24.15% | ~ |
| `Compile/fields1` | +1.52% | +1.05% | ~ |

`allocs/op` is **unchanged** — this is *bigger* allocations, not *more*. A `string` header is 16 B with no separate allocation (backing bytes are substrings of the already-allocated pattern), so it inflates `B/op` (scaling with field count) without touching `allocs/op`. Absolute cost is tiny (`CompilePlanOnly` 294→317 ns) and paid once per `Compile`.

### 3. Conversion-error path — intended cost of `*DecodeError`
| benchmark | sec/op | B/op | allocs/op |
|---|---|---|---|
| `Unmarshal/invalidIntConversion` | +11.08% | +21.91% (881→1074 B) | **+2** (16→18) |
| `Unmarshal/unsupportedFieldType` | +16.64% | +18.73% (598→710 B) | **+2** (10→12) |

This is exactly the new `*DecodeError{Field,Group,Value,Type,Err}` being heap-allocated with its richer message. Only on the **failure** path (bad input), which isn't hot.

### Consumer impact
None of concern. The success path consumers actually run repeatedly is allocation-neutral and time-flat. `Compile` cost amortizes under compile-once-reuse. The +2 allocs only appear on a failed conversion. Error inspection is now via the durable typed-error/sentinel contract (`errors.As`/`errors.Is` on `*DecodeError` / `ErrNoMatch`) — message wording is not a supported dependency, so the prefix/wording changes are not a back-compat concern.

<sub>Raw `benchstat` data captured locally (`-count=10`, full base suite + focused head set).</sub>

decoder.go
```diff
@@ -233,6 +233,26 @@ func matchGroupName(re *regexp.Regexp, fieldName string) string {
...
+// occurrence of a reused name resolves to the same string. The empty case — a
+// default-only field with no declared group — is reached only when a `default=`
+// value itself fails to convert on the lenient Unmarshal path; there the name is
+// the explicit tag (or "" if untagged), worth re-parsing the tag for in that
+// rare case. This matches the name buildDecodePlan resolved at build time.
+func resolveGroupName(re *regexp.Regexp, sf reflect.StructField, groupIndexes []int) string {
+	if len(groupIndexes) > 0 {
+		return re.SubexpNames()[groupIndexes[0]]
+	}
+	name, _, _ := parseFieldTag(sf)
```
Jecoms, line 252: `[dust]` This `parseFieldTag` fallback branch is the one the lazy refactor changed, and the perf commit's "byte-for-byte identical Group" claim rests on it. It *executes* (the `malformed default value` test at unmarshal_test.go drives a tagged→undeclared-group `default=` failure, so `resolveGroupName` shows 100% func coverage), but no test asserts the resulting `de.Group` for the fallback cases — tagged→undeclared `default=` should yield the tag name, empty-name default-only should yield `""`. I verified both manually; a one-line `de.Group` assertion on the existing default-failure test would lock the contract against future drift. (commit f4b53e5; resolveGroupName godoc)

Jecoms (commented):
### Code review

_Re-review focused on the two commits added since the prior approve-with-nits: `e5ae79c` (merge of #143, re-deriving #111 onto the unified decode path) and `f4b53e5` (lazy `DecodeError.Group` resolution)._

**2 finding(s)** — 2 dust. Inline + PR-level below.

What I verified independently (all green):
- Build `go build ./...`, `go test -race ./...`, `gofmt -s -l .`, `go vet ./...`, `golangci-lint run` (0 issues) — all pass locally; CI checks all pass; coverage 97.8% (> 92% gate).
- **`DecodeError.Group` is byte-for-byte identical after the lazy refactor** — confirmed empirically for all four cases: tagged→declared group (`"age"`), untagged resolved by Unicode fold (`"count"`), **tagged→undeclared group with `default=` → tag name** (`"missing"`), and **empty-name default-only → `""`**. The identity holds because tag→group matching is exact, so `re.SubexpNames()[groupIndexes[0]]` equals the verbatim tag/declared name; the `parseFieldTag` fallback only runs when `groupIndexes` is empty.
- **No panic risk** in `re.SubexpNames()[groupIndexes[0]]`: every entry in `groupIndexes` is an index produced by iterating `re.SubexpNames()` (in `subexpIndexes`), and it's read only under `len(groupIndexes) > 0`.
- `re` threaded correctly into all three `runDecodePlan` callers (`Decoder.decode` via `d.re`, `Unmarshal`, `UnmarshalAll`).
- All five entrypoint prefixes correct (`regextra.Unmarshal:`, `regextra.UnmarshalAll: match N:`, `regextra.Decoder.One:`/`All:`/`Iter:`); `ErrNoMatch` keeps the bare `regextra:` prefix.
- `DecodeError.Group` godoc (strict-vs-lenient framing) and the CHANGELOG #108/#111 reconciliation are accurate against the final merged state; no conflict markers remain.

No blocking issues — the re-derivation and perf refactor are correct.

#### PR-level

1. `[dust]` `CHANGELOG.md` #111 Added entry says "All three entrypoints decode through one shared core" but the same sentence lists five (`Unmarshal`, `UnmarshalAll`, `Decoder.One`/`All`/`Iter`). Minor wording — "three" reads as the three API groupings but is easy to misread. (CHANGELOG.md, Unreleased › Added › #111)

decoder.go
Jecoms: Addressed in 9ec11dd — added two subtests in `TestUnmarshalDecodeError` that pin `DecodeError.Group` for both shapes of the `resolveGroupName` default-only fallback (the `groupIndexes`-empty branch): an untagged field resolves `Group` to `""`, and a field tagged to an undeclared group resolves `Group` to the tag name. Coverage 97.8% → 98.1%.

Jecoms:
Addressed the two re-review dust nits in 9ec11dd:

- **`resolveGroupName` fallback test** (inline thread) — added assertions pinning `DecodeError.Group` for both default-only fallback shapes.
- **CHANGELOG #111 wording** (PR-level) — reworded "All three entrypoints decode through one shared core" to "They all decode through one shared core", dropping the three-vs-five mismatch.

Gate stays green: `go build`/`vet`/`gofmt` clean, `go test -race` ok, golangci-lint 0 issues, coverage 98.1%.

Jecoms:
@coderabbitai addressed your nitpick (from the `e5ae79c` review) in 992c8ff: the forward-safety `buildDecodePlan` error branches in `Unmarshal`/`UnmarshalAll` now wrap with `fmt.Errorf("regextra.Unmarshal: %w", err)` / `"regextra.UnmarshalAll: %w"` so they honor the entrypoint prefix contract if that branch ever becomes reachable. Gate stays green (build/vet/gofmt, `go test -race`, golangci-lint 0).

PR #143, opened 2026-06-30
refactor: route Unmarshal/UnmarshalAll through the shared decode plan

### LINKED ISSUE #108: Refactor: route Unmarshal through a single shared decode plan (unify Unmarshal/Decoder semantics)

Jecoms:
## Summary

- Extracts the `Decoder`'s plan build/run into two non-generic free functions — `buildDecodePlan(rt, re, strict)` and `runDecodePlan(fields, rv, target, matches)` — and routes `Unmarshal`/`UnmarshalAll` through them with `strict=false`. `compileDecoder` delegates with `strict=true`; `Decoder.decode` is now a thin `runDecodePlan` wrapper. One decode core and one set of field-mapping/skip-or-default semantics for all three paths, so they can't drift again (the root cause behind #104/#105/#106).
- `UnmarshalAll` builds the plan once per call and runs it per match into a pre-sized slice, removing the per-match `map[string]string` build and per-field tag re-parse the old `populateStruct` loop paid on every match. Deletes the now-dead `populateStruct` and `findGroupValue`.
- The `strict` flag preserves the behavioral split: `Decoder` validates eagerly (undeclared group, bad `default=`, misplaced `layout=`); `Unmarshal` stays lenient.

### Intentional, non-breaking nuances (no public API change)
- The wrapping prefix on a field-conversion error from `Unmarshal`/`UnmarshalAll` changes from `regextra: failed to set field X: …` to `field X: …` (matching `Decoder`; the underlying `cannot convert …` message is unchanged; no test asserted the old prefix).
- When several declared group names fold-equal one untagged field, the field-name fallback now resolves deterministically in pattern-declaration order (`matchGroupName`) instead of the old map-iteration order — replacing previously-unstable behavior.

Both are recorded in the CHANGELOG.

## Test plan
- [x] `go build ./...`, `go vet ./...`, `gofmt -s -l .` clean
- [x] `go test -race ./...` green; coverage 97.4% (≥92% gate)
- [x] Parity tests stay green (`TestUnmarshal_decoderParityOnOptionalGroups`, `TestCaseInsensitiveFallback_parity`, duplicate-name + optional-group cases)
- [x] `BenchmarkUnmarshalAll` before/after: `matches100` 408→117 allocs/op (~37% faster), `matches1000` 4011→1020 allocs/op, `realisticLogCorpus` 308→114 allocs/op. Single-match wall-time flat/slightly better; alloc count ticks up modestly (one-shot path now builds an uncached plan, as designed) — amortized away on the multi-match path.

Fixes #108


Jecoms (commented):
### Code review

**6 finding(s)** — 2 pebble, 3 sand, 1 dust. Inline comments below.

Clean refactor: two independent reviewers verified the new shared-plan path is behavior-equivalent to the deleted `populateStruct`/`findGroupValue`, the #104/#105/#106 skip-or-default and last-participating-occurrence contracts are preserved, and `buildDecodePlan` never errors when `strict=false`. No bugs found. The findings below are about documentation/test completeness and convention consistency, not correctness.

#### PR-level

1. `[sand]` PR is typed `refactor:`, which CONTRIBUTING.md defines as "Code restructuring without behavior or API change," yet it introduces two documented intentional behavior nuances (the field-conversion error prefix `regextra: failed to set field X:` → `field X:`, and deterministic fold-collision resolution). Non-breaking with no public-API/contract change, so `refactor` is defensible, but the type understates the observable changes. (CONTRIBUTING.md commit-type table)

   https://github.com/Jecoms/regextra/blob/15e3b3c2d5b0faa6d563bd7be873bcd7e5055c57/CHANGELOG.md#L14-L16

2. `[dust]` The error-prefix change is back-compat-sensitive (a downstream `strings.Contains` on the old text would break); it's sanctioned non-breaking and CHANGELOG-documented. Concurrent PR #142 rewrites the same `field %s: %w` wrap — watch for merge drift between the two. (prior-PR #142 review)

   https://github.com/Jecoms/regextra/blob/15e3b3c2d5b0faa6d563bd7be873bcd7e5055c57/decoder.go#L363-L366

decoder.go
```diff
@@ -117,42 +147,44 @@ func compileDecoder[T any](pattern string, re *regexp.Regexp) (*Decoder[T], erro
 		if groupName == "" {
 			// No explicit tag — fall back to matching the field name against a
 			// declared group: exact first, then case-insensitively via Unicode
-			// simple-fold (the same predicate Unmarshal's runtime fallback
-			// uses; see matchGroupName). A field that matches no group and has
-			// no default is treated as a typo and fails at compile time below.
+			// simple-fold (see matchGroupName). A field that matches no group
+			// and has no default is treated as a typo and, under strict, fails
+			// the build below.
 			groupName = matchGroupName(re, sf.Name)
```
Jecoms, line 153: `[pebble]` The CHANGELOG entry frames the fold-collision fallback change as only "declaration order instead of map-iteration order," but moving the untagged-field fallback to build time via `matchGroupName` (participation-agnostic, over `re.SubexpNames()`) is also a participation change: when two distinctly-spelled declared names fold-equal one untagged field and the declaration-first one does not participate while a fold-sibling does, this now skips the field where the old participating-only fallback (over `namedGroupValues(...,false)`) populated it. Extremely obscure and it aligns Unmarshal with Decoder (the goal), so defensible — but worth documenting the participation aspect (and ideally pinning with a test). (CHANGELOG #108 entry; corroborated against `findGroupValue` history f9b9a97/76f4f68)

decoder.go
```diff
@@ -284,14 +316,23 @@ func (d *Decoder[T]) Pattern() string {
...
+	return runDecodePlan(d.fields, rv, target, matches)
+}
+
+// runDecodePlan executes a decode plan (from [buildDecodePlan]) against a single
+// match's group indices, writing the decoded values into rv — the addressable
+// reflect.Value of a struct. matches is a FindStringSubmatchIndex-style index
+// slice (or one element of FindAllStringSubmatchIndex); target is the string
+// those indices slice into. It is the single decode core shared by [Decoder]
+// (One/All/Iter) and the [Unmarshal] / [UnmarshalAll] free functions.
+func runDecodePlan(fields []fieldDecoder, rv reflect.Value, target string, matches []int) error {
```
Jecoms, line 333: `[pebble]` This refactor extracts the Decoder's compile path (`compileDecoder` → `buildDecodePlan`) and the shared One/All/Iter decode core (`Decoder.decode` → `runDecodePlan`), but adds no new Decoder/Iter or Unmarshal test — nor one pinning the two newly-documented nuances (error prefix, deterministic fold order). Prior reviews on these files (#123/#127/#132/#134) repeatedly asked for an explicit Decoder/Iter case when a compile/decode path changes; a regression test would lock the extracted path against future drift. (CONTRIBUTING.md "Test the public contract"; prior-PR pattern)

regextra.go
```diff
@@ -391,14 +390,15 @@ func NamedGroupsPerMatchSeq(re *regexp.Regexp, target string) iter.Seq[map[strin
...
-//     declared group.
-//   - false (the [Unmarshal] / [UnmarshalAll] path): the name is omitted, so
-//     a non-participating optional group looks like "no value" to
-//     populateStruct — it is left at its zero value rather than handed "",
-//     which would fail conversion on a typed field. A duplicate that
-//     participates elsewhere still sets the key, so omitting here never drops
-//     a real value.
+//   - true (the [NamedGroups] family — its only callers today): the name is
+//     recorded as "" so callers see every declared group.
+//   - false: the name is omitted entirely, so a non-participating optional
```
Jecoms, line 395: `[sand]` After routing the typed path off this map, no caller passes `includeNonParticipating=false` anymore (all three remaining callers — the `NamedGroups` family — pass `true`), so this branch is now dead. The godoc honestly documents this and keeps the flag "for the omit-vs-empty distinction," so it's defensible; but consider dropping the now-unreachable branch to keep the function focused. (CONTRIBUTING.md "prefer simplicity over cleverness")

decoder.go
```diff
@@ -284,14 +316,23 @@ func (d *Decoder[T]) Pattern() string {
...
 }
 
-// decode walks the precomputed field plan against a single match's group
-// indices and writes the values into rv (the addressable reflect.Value of a
-// T). matches is a FindStringSubmatchIndex-style index slice (or one element
-// of FindAllStringSubmatchIndex); target is the string those indices slice
-// into.
+// decode walks the precomputed field plan against a single match and writes the
+// values into rv (the addressable reflect.Value of a T). It is a thin wrapper
+// over the shared [runDecodePlan] core, which the [Unmarshal] / [UnmarshalAll]
```
Jecoms: `[sand]` The new godoc doc-links unexported symbols (`[runDecodePlan]`, `[buildDecodePlan]`); pkg.go.dev / `go doc` degrade brackets on unexported names to plain text. Prior reviews (#85/#134) flagged this exact pattern, and the same PR's `namedGroupValues` edit uses unbracketed `buildDecodePlan / runDecodePlan` — so the two files are now inconsistent. (prior-PR convention #85/#134)

decoder.go
Jecoms: Addressed in 27ee857. Added `TestUnmarshalDecoder_foldFallbackDeterministicParticipation`, which pins both the deterministic declaration-order fold fallback and the participation nuance you flagged: with `(?:a(?P<value>\d+)|b(?P<VALUE>\d+))` and an untagged `Value` field, `matchGroupName` binds the declaration-first `value` participation-agnostically, so input `b5` (only `VALUE` participates) now leaves the field at zero on **both** Unmarshal and Decoder — where the old participating-only fallback populated it. The test asserts the two paths agree on `a5`/`b5`. Also rewrote the CHANGELOG #108 nuance (2) to spell out the participation aspect (build-time fallback over `re.SubexpNames()` is participation-agnostic), not just declaration- vs map-iteration order.

decoder.go
Jecoms: Addressed in 27ee857. Added `TestSharedDecodePlan_AllPathsAgree`, which exercises the extracted `buildDecodePlan`/`runDecodePlan` core through one struct touching a default, a duplicate group name (last-participating-occurrence wins), and a typed conversion, then asserts every path that runs the shared core — Decoder.One/All/Iter and Unmarshal/UnmarshalAll — decodes each match identically. The two newly-documented nuances are pinned by the sibling tests added in the same commit: `TestUnmarshalDecoder_fieldConversionErrorPrefixUnified` (the `field X:` prefix unification across Unmarshal/UnmarshalAll/Decoder.One) and `TestUnmarshalDecoder_foldFallbackDeterministicParticipation` (deterministic fold order). That covers the explicit Decoder/Iter + nuance regression the #123/#127/#132/#134 reviews kept asking for.

regextra.go
Jecoms: Leaving as-is. The `includeNonParticipating=false` branch is intentionally retained per the issue-#108 plan: it preserves the omit-vs-empty distinction (`NamedGroups` family passes `true` and records non-participating groups as `""`; the `false` mode omits them so a non-participating optional group reads as "no value"). The godoc already documents that the typed path no longer routes through this map but the flag is kept for that distinction. Deleting the branch would erase a documented, deliberately-preserved capability rather than dead-code cleanup, so I'd rather not narrow it in this refactor. Happy to file a separate issue if you'd prefer to revisit the omit-vs-empty surface on its own.

decoder.go
Jecoms: Fixed in 27ee857. Unbracketed the unexported doc-links in decoder.go (`buildDecodePlan`/`runDecodePlan`) so they're plain text, matching the same PR's `namedGroupValues` godoc and the #85/#134 convention. The exported `[Unmarshal]`/`[UnmarshalAll]` links on those lines stay bracketed.

Jecoms:
Follow-up in 27ee857 addressing the inline findings (replies on each thread):

- **[pebble] decoder.go:153** — added `TestUnmarshalDecoder_foldFallbackDeterministicParticipation` pinning the deterministic fold order **and** the participation nuance; rewrote the CHANGELOG #108 nuance (2) to call out the participation aspect.
- **[pebble] decoder.go:333** — added `TestSharedDecodePlan_AllPathsAgree` (extracted `buildDecodePlan`/`runDecodePlan` core across Decoder.One/All/Iter + Unmarshal/UnmarshalAll), plus `TestUnmarshalDecoder_fieldConversionErrorPrefixUnified` for the error-prefix nuance.
- **[sand] decoder.go:321** — unbracketed the unexported `buildDecodePlan`/`runDecodePlan` doc-links to match the #85/#134 convention.
- **[sand] regextra.go:395** — left as-is (the `includeNonParticipating=false` branch is the deliberately-retained omit-vs-empty distinction per the #108 plan); rationale on the thread.

On the two PR-level findings (no inline thread to reply on):

- **[sand] `refactor:` type vs documented nuances** — acknowledged. Keeping the `refactor:` type: there's no public-API or contract change, and both nuances are non-breaking and CHANGELOG-documented under the sanctioned error-wording carve-out. The CHANGELOG now states them precisely; treating this as out-of-scope for a commit-type change.
- **[dust] merge-drift vs concurrent PR #142** — acknowledged and noted. Out of scope here: #142 owns the `field %s: %w` wrap rewrite, so I'm deliberately not touching that area from this PR. Whichever lands second resolves the wrap; the new error-prefix test guards the observable `field X:` contract on this side.

Verification (from the worktree): `go build ./...`, `go vet ./...`, `go test -race ./...`, `gofmt -s -l .` (clean), `golangci-lint run` (0 issues) all pass.

PR #139, opened 2026-06-30
test: consolidate test/bench files to one sibling file per source

Jecoms:
## Summary
- Mechanical move-and-merge of the fragmented test/benchmark suite into Go's idiomatic one-sibling-file-per-source layout — **no test or benchmark logic changed**. Every `Test*`/`Benchmark*` is routed to the sibling of the source it exercises by function-name prefix.
  - Tests merged into `regextra_test.go`, `unmarshal_test.go`, `decoder_test.go`.
  - Benches renamed/merged into `regextra_bench_test.go`, `unmarshal_bench_test.go`, `decoder_bench_test.go` (the `_bench_test.go` suffix sorts each next to its source + test).
  - `bench_internal_test.go` (`package regextra`, needs unexported `compileDecoder`) and `bench_sanity_test.go` (cross-cutting fixture sanity) kept as the package-scoped exceptions.
- Cross-cutting topical files (`casefold`, `duplicate_groups`, `unmarshal_exclude/optional/overflow/tag`) were **split by func-name prefix**, not flat-moved. The four internal-package files were converted to `package regextra_test` (verified none reference unexported symbols) and their calls qualified with the `rx` import alias.
- Hardening to prevent recurrence (no CI check, per the issue): documented the one-file-per-source rule in `CONTRIBUTING.md` and refreshed its stale `## Project Structure` section (it still described a `main.go`/`main_test.go` layout); added `AGENTS.md` (thin pointer to `CONTRIBUTING.md`) and `CLAUDE.md` (pointer to `AGENTS.md`).

All 103 top-level `Test`/`Example`/`Fuzz`/`Benchmark` functions are preserved (verified by diffing the function-name set against `origin/main`).

## Test plan
- [x] `go test -race ./...` passes; coverage **97.0%** (gate is 92.0%).
- [x] `go test -run x -bench . -benchtime=1x ./...` — all benchmarks compile and run.
- [x] `gofmt -s -l .` clean, `go vet ./...` clean, `golangci-lint run` → 0 issues.
- [x] Top-level test/bench/example function set is byte-identical to `origin/main` (none added or lost).

Fixes #138

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Added a new agent-facing guide that points to the main contributor guide and highlights key repository conventions.
  * Updated agent guidance to reference the preferred source for contribution instructions.
  * Expanded contributor documentation with clearer rules for test and benchmark placement, plus an updated project structure overview.

PR #157, opened 2026-06-30
test: cover remaining low-coverage decode-plan branches

### LINKED ISSUE #115: Tests: cover remaining low-coverage branches
Coverage is 94.6% overall but a few functions lag, all on the same kinds of paths:

- `matchGroupName` (**71.4%**): exact-match branch (decoder.go:167) and no-match branch (decoder.go:176) — every existing test happens to hit the case-insensitive path
- `findGroupValue` (**80.0%**): exact-case field-name match (unmarshal.go:284) and miss-everything return (unmarshal.go:296)
- `setFieldValue` unsupported-kind error (unmarshal.go:434): a struct/slice/map field with a matching group is a plausible user mistake; zero coverage. Also document in the `Unmarshal` godoc that nested structs are not flattened and produce this error.
- `UnmarshalAll` conversion-error mid-iteration (unmarshal.go:169) — the `Decoder.All` equivalent IS tested, this one isn't
- Decoder compile skips: unexported field (decoder.go:101), untagged field with no matching group (decoder.go:149)

## Implementation plan

**Status note — the issue body is partly stale.** Total coverage is now **98.3%** (not 94.6%), and the #143 refactor ("route Unmarshal/UnmarshalAll through the shared decode plan") changed the map. `findGroupValue` no longer exists — its exact-case match folded into `matchGroupName` and its miss-return into the shared plan. Several originally-listed branches are now covered: `matchGroupName`'s no-match return, the unexported-field skip, and `UnmarshalAll`'s mid-iteration conversion error all show as covered in the current profile. The line numbers below are current.

**Approach.** Pure test additions (one godoc edit) — no behavior change. Add targeted table cases that exercise the few branches a coverage run still flags, then confirm the profile. Nothing here requires new test seams.

**Coverable branches remaining** (from `go tool cover`, count 0):
- `matchGroupName` exact-match branch — `decoder.go:225` (`re.SubexpIndex(fieldName) != -1`). Untagged field whose Go name equals a declared group exactly (no case-folding). e.g. group `(?P<Name>...)` + field `Name` with no tag.
- `buildDecodePlan` untagged-no-match skip — `decoder.go:193`. Untagged field whose name matches no group and has no `default=` → silently skipped (Unmarshal/lenient path; under strict Compile it errors instead, so drive this via `Unmarshal`).
- `setFieldValue` unsupported-kind error — `unmarshal.go:503` (`default:` arm). A struct/slice/map field bound to a matching group → `unsupported field type`. **Also** add a line to the `Unmarshal` godoc "Returns an error if" list: nested struct/slice/map fields are not flattened and produce this error.
- (bonus, not in original list) `encoding.TextUnmarshaler` interface-path error — `unmarshal.go:455`. An interface-typed field holding a TextUnmarshaler whose `UnmarshalText` returns an error.

**Intentionally NOT coverable — do not chase.** `unmarshal.go:159` and `unmarshal.go:221` are the `if err := buildDecodePlan(...); err != nil` guards on the lenient path. `buildDecodePlan(strict=false)` never returns a non-nil error (documented at its definition and at each call site as forward-safety). Leave them; note them in the PR description so a reviewer doesn't expect 100%.

**Files / surfaces touched.**
- `regextra_test.go` — new table cases for the four branches above.
- `unmarshal.go` — one-line godoc addition to `Unmarshal` (nested struct/slice/map → error).

**Verification.**
- `go test ./...` green.
- `go test -coverprofile=cov.out ./... && go tool cover -func=cov.out` — the four target blocks no longer report 0; only the two forward-safety guards remain uncovered.

**Open questions / risks.**
- Per CONTRIBUTING.md, confirm no benchmark validation is needed (test-only change, no hot-path edit).

Jecoms:
## Summary
Pure test additions (plus one godoc line) closing the coverage gaps called out in #115. No behavior change. Total coverage 98.3% → 99.2%.

Branches now covered (all confirmed via `go tool cover -func`, previously count 0):
- `matchGroupName` exact-match arm (`decoder.go`) — untagged field whose Go name equals a declared group *exactly* (no case-fold). → `TestCompile_fieldNameExactMatch`
- `buildDecodePlan` untagged-no-match skip (`decoder.go`) — untagged field matching no group and with no `default=` is silently skipped and left at zero value. → `TestUnmarshal_untaggedFieldNoMatchingGroupSkipped`
- `setFieldValue` unsupported-kind arm (`unmarshal.go`) — a struct/slice/map field bound to a matching group yields `unsupported field type`. → `TestUnmarshal_unsupportedFieldType`
- `encoding.TextUnmarshaler` interface-path error (`unmarshal.go`) — interface-typed field whose `UnmarshalText` returns an error (distinct from the already-covered addressable arm). → `TestUnmarshal_TextUnmarshaler_InterfaceField_Error`

Also documents in the `Unmarshal` godoc that nested struct/slice/map fields are not flattened and produce an `unsupported field type` error.

### Notes for the reviewer
- Two blocks remain intentionally uncovered: the `if err := buildDecodePlan(...); err != nil` forward-safety guards on the lenient `Unmarshal`/`UnmarshalAll` path (`unmarshal.go`). `buildDecodePlan(strict=false)` never returns a non-nil error, so these are unreachable by design — not chased, per the issue's implementation plan.
- Tests are routed per source file per CONTRIBUTING.md (`TestCompile*` → `decoder_test.go`, `TestUnmarshal*` → `unmarshal_test.go`), rather than the plan's `regextra_test.go` suggestion.
- Plan nuance: the plan said the untagged-no-match case "errors under strict Compile"; in current code `matchGroupName` returns `""`, so the strict missing-group error is bypassed and the field is skipped under both paths. The test drives it via `Unmarshal` as planned and asserts the skip.
- Test-only change, no hot-path edit → no benchmark validation required (CONTRIBUTING.md).

## Test plan
- [x] `go test ./...` — green
- [x] `go vet ./...` — clean
- [x] `go test -coverprofile=cov.out ./... && go tool cover -func=cov.out` — the four target blocks report 100%; only the two forward-safety guards remain uncovered
- [x] `gofmt -l` — clean

Fixes #115

### DIFF
--- a/unmarshal.go
+++ b/unmarshal.go
@@ -119,6 +119,10 @@ func (e *DecodeError) Unwrap() error { return e.Err }
 // Returns an error if:
 //   - v is not a pointer to a struct
 //   - Type conversion fails on a matched group
+//   - A matched field is a nested struct, slice, or map: these are not
+//     flattened, so a group bound to one yields an "unsupported field type"
+//     error (unless the type implements [RegexUnmarshaler] or
+//     encoding.TextUnmarshaler, which convert themselves)
 //
 // Example:
 //
--- a/unmarshal_test.go
+++ b/unmarshal_test.go
@@ -1837,6 +1837,26 @@ func TestUnmarshal_TextUnmarshaler_Error(t *testing.T) {
 	}
 }
 
+// TestUnmarshal_TextUnmarshaler_Value covers the addressable (CanAddr) error
+// arm. An interface-typed field reaches UnmarshalText via Type().Implements
+// instead, so its error arm is distinct — drive it with the same bad input.
+func TestUnmarshal_TextUnmarshaler_InterfaceField_Error(t *testing.T) {
+	type rec struct {
+		Addr encoding.TextUnmarshaler `regex:"addr"`
+	}
+
+	re := regexp.MustCompile(`(?P<addr>\S+)`)
+	var ip netip.Addr
+	got := rec{Addr: &ip}
+	err := rx.Unmarshal(re, "not-an-ip", &got)
+	if err == nil {
+		t.Fatal("expected error for invalid netip.Addr via interface field, got nil")
+	}
+	if !strings.Contains(err.Error(), "not-an-ip") {
+		t.Errorf("error %q should mention the offending value", err)
+	}
+}
+
 // regexAndText implements BOTH RegexUnmarshaler and encoding.TextUnmarshaler;
 // RegexUnmarshaler must win.
 type regexAndText struct {
@@ -2023,5 +2043,48 @@ func TestUnmarshalArgErrorPrefixes(t *testing.T) {
 	if err := rx.UnmarshalAll(re, "30", nil); err == nil ||
 		!strings.HasPrefix(err.Error(), "regextra.UnmarshalAll:") {
 		t.Errorf("UnmarshalAll(nil) error = %v, want regextra.UnmarshalAll: prefix", err)
+	}
+}
+
+// A field whose kind setFieldValue can't convert (here a nested struct, but a
+// slice or map behaves the same) and which doesn't implement RegexUnmarshaler
+// or encoding.TextUnmarshaler falls through to the "unsupported field type"
+// arm. Nested structs are not flattened — a plausible user mistake.
+func TestUnmarshal_unsupportedFieldType(t *testing.T) {
+	type Nested struct{ X string }
+	type rec struct {
+		Field Nested `regex:"field"`
+	}
+
+	re := regexp.MustCompile(`(?P<field>\w+)`)
+	var got rec
+	err := rx.Unmarshal(re, "hello", &got)
+	if err == nil {
+		t.Fatal("expected an error for a struct-typed field, got nil")
+	}
+	if !strings.Contains(err.Error(), "unsupported field type") {
+		t.Errorf("error %q, want it to mention %q", err, "unsupported field type")
+	}
+}
+
+// An untagged exported field whose name matches no declared group, with no
+// default= to fall back on, is silently skipped by the lenient decode plan
+// (decoder.go) and left at its zero value rather than erroring.
+func TestUnmarshal_untaggedFieldNoMatchingGroupSkipped(t *testing.T) {
+	type rec struct {
+		Name    string `regex:"name"`
+		Missing string // no tag, no "missing"/"Missing" group, no default — skipped
+	}
+
+	re := regexp.MustCompile(`(?P<name>\w+)`)
+	got := rec{Missing: "untouched"}
+	if err := rx.Unmarshal(re, "Alice", &got); err != nil {
+		t.Fatalf("Unmarshal() error = %v, want nil (unmapped field should be skipped)", err)
+	}
+	if got.Name != "Alice" {
+		t.Errorf("Name = %q, want %q", got.Name, "Alice")
+	}
+	if got.Missing != "untouched" {
+		t.Errorf("Missing = %q, want it left unchanged at %q", got.Missing, "untouched")
 	}
 }
--- a/unmarshal_test.go
+++ b/unmarshal_test.go
@@ -1837,8 +1837,8 @@ func TestUnmarshal_TextUnmarshaler_Error(t *testing.T) {
 	}
 }
 
-// TestUnmarshal_TextUnmarshaler_Value covers the addressable (CanAddr) error
-// arm. An interface-typed field reaches UnmarshalText via Type().Implements
+// TestUnmarshal_TextUnmarshaler_Error (above) covers the addressable (CanAddr)
+// error arm. An interface-typed field reaches UnmarshalText via Type().Implements
 // instead, so its error arm is distinct — drive it with the same bad input.
 func TestUnmarshal_TextUnmarshaler_InterfaceField_Error(t *testing.T) {
 	type rec struct {
@@ -1852,8 +1852,15 @@ func TestUnmarshal_TextUnmarshaler_InterfaceField_Error(t *testing.T) {
 	if err == nil {
 		t.Fatal("expected error for invalid netip.Addr via interface field, got nil")
 	}
-	if !strings.Contains(err.Error(), "not-an-ip") {
-		t.Errorf("error %q should mention the offending value", err)
+	var de *rx.DecodeError
+	if !errors.As(err, &de) {
+		t.Fatalf("error %q is not a *DecodeError", err)
+	}
+	if de.Field != "Addr" {
+		t.Errorf("DecodeError.Field = %q, want %q", de.Field, "Addr")
+	}
+	if de.Value != "not-an-ip" {
+		t.Errorf("DecodeError.Value = %q, want %q", de.Value, "not-an-ip")
 	}
 }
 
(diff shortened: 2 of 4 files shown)


Jecoms (commented):
### Code review

**5 finding(s)** — 1 boulder, 1 pebble, 2 sand, 1 dust. Inline comments below.

Clean, well-scoped coverage PR; the body pre-empts the routing/benchmark/coverage questions and the new tests genuinely exercise their target branches (verified independently). One blocking item is process, not code.

#### PR-level

1. `[boulder]` The single commit `0fc305a` is **unsigned / unverified** (`gh api .../pulls/157/commits` → `verification.reason: "unsigned"`). This repo requires all PR commits to be verified (signed) before merge — re-sign the commit (amend/rebase with your signing key) and force-push. (Source: maintainer signing policy — `signing.sign-all`; CLAUDE.md "All commits in PRs MUST be verified".)

unmarshal_test.go
```diff
@@ -1837,6 +1837,26 @@ func TestUnmarshal_TextUnmarshaler_Error(t *testing.T) {
 	}
 }
 
+// TestUnmarshal_TextUnmarshaler_Value covers the addressable (CanAddr) error
```
Jecoms: `[pebble]` This comment misattributes the sibling test: `TestUnmarshal_TextUnmarshaler_Value` (line 1757) is a **success**-path test and covers no error arm. The addressable (CanAddr) **error** arm is actually covered by `TestUnmarshal_TextUnmarshaler_Error` (line 1824) — the test directly above. Point the comment at `..._Error` so the contrast it's drawing is accurate. (Independently flagged by 3 reviewers; echoes prior comment-accuracy feedback on #139/"don't claim what you don't assert" #134.)

unmarshal_test.go
```diff
@@ -2025,3 +2045,46 @@ func TestUnmarshalArgErrorPrefixes(t *testing.T) {
 		t.Errorf("UnmarshalAll(nil) error = %v, want regextra.UnmarshalAll: prefix", err)
 	}
 }
+
+// A field whose kind setFieldValue can't convert (here a nested struct, but a
+// slice or map behaves the same) and which doesn't implement RegexUnmarshaler
+// or encoding.TextUnmarshaler falls through to the "unsupported field type"
+// arm. Nested structs are not flattened — a plausible user mistake.
+func TestUnmarshal_unsupportedFieldType(t *testing.T) {
```
Jecoms, line 2060: `[sand]` Both `TestUnmarshal_unsupportedFieldType` (here) and `TestUnmarshal_untaggedFieldNoMatchingGroupSkipped` (line 2073) drive shared decode-plan branches (`setFieldValue`, `buildDecodePlan`/`matchGroupName`) only through the reflect `Unmarshal` path. Prior reviews (#123/#127/#132/#134/#143) established a pattern of adding a sibling `Compile`/`Decoder.One/Iter` case when a shared compile/decode branch is exercised — `TestCompile_fieldNameExactMatch` already follows it. Consider mirroring these two on the typed path. (Source: recurring reviewer pattern, prior PRs.)

unmarshal_test.go
```diff
@@ -2025,3 +2045,46 @@ func TestUnmarshalArgErrorPrefixes(t *testing.T) {
...
+		Field Nested `regex:"field"`
+	}
+
+	re := regexp.MustCompile(`(?P<field>\w+)`)
+	var got rec
+	err := rx.Unmarshal(re, "hello", &got)
+	if err == nil {
+		t.Fatal("expected an error for a struct-typed field, got nil")
+	}
+	if !strings.Contains(err.Error(), "unsupported field type") {
```
Jecoms: `[sand]` This (and the `"not-an-ip"` check at line 1855) asserts on `err.Error()` substrings. The repo's error contract is typed errors (`errors.As`/`errors.Is`, `DecodeError`), not message wording — #142 added `errors.As`/`DecodeError.Group` assertions for exactly this reason. Consider `errors.As(err, &de)` plus a field assertion so the test locks the contract rather than the message text. (Source: typed-error contract; prior PR #142.)

unmarshal.go
```diff
@@ -119,6 +119,10 @@ func (e *DecodeError) Unwrap() error { return e.Err }
 // Returns an error if:
 //   - v is not a pointer to a struct
 //   - Type conversion fails on a matched group
+//   - A matched field is a nested struct, slice, or map: these are not
+//     flattened, so a group bound to one yields an "unsupported field type"
```
Jecoms, line 123: `[dust]` The new "unsupported field type" / not-flattened note is accurate for `Unmarshal`, but the same arm is reachable via `Compile`/`Decoder`. For doc consistency across both entry points you may want to mirror this note on the typed-path godoc (non-blocking). (Source: cross-path godoc consistency, echoes #142.)

unmarshal_test.go
Jecoms: Fixed in e506b30 — the comment now credits `TestUnmarshal_TextUnmarshaler_Error` (the addressable CanAddr error arm directly above), not the success-path `..._Value` test.

unmarshal_test.go
Jecoms: Addressed in e506b30 — added typed-path siblings in `decoder_test.go`: `TestDecoder_unsupportedFieldType` (Compile succeeds, `One` surfaces the `*DecodeError`) and `TestDecoder_untaggedFieldNoMatchingGroupSkipped`. I verified both behaviors mirror 1:1 on the strict path — only a *tagged* missing-group reference fails Compile; an untagged no-group field is skipped on both paths.

unmarshal_test.go
Jecoms: Addressed in e506b30 — both this test and the new `..._InterfaceField_Error` now assert via `errors.As(err, &de)` on `DecodeError.Field`/`Group`/`Value`, matching the #142 typed-error contract. Kept one `de.Unwrap()` substring check only to pin the "unsupported field type" *arm* (the typed fields don't otherwise distinguish it from a convert failure). Left the pre-existing `..._Error` at line 1835 as-is — out of this PR's diff scope.

unmarshal.go
Jecoms: Addressed in e506b30 — mirrored the note on the typed path. Placed it on `Decoder.One`'s godoc (the decode-time entry where the arm actually surfaces — `Compile` validates tags, not field kinds) and noted it applies to `All`/`Iter` too, since they share One's decode path.

Jecoms:
Addressed review feedback in e506b30 (force-pushed; both commits now show **Verified**):

- **`[boulder]` unsigned commit** — re-signed the base commit (now `87e7a77`, Verified) and the new fix commit (`e506b30`, Verified). Separately, this signing requirement only lived in a personal gitignored `CLAUDE.md`; I've moved it into the tracked source of truth (`CONTRIBUTING.md`) in #158 so the policy is discoverable and authoritative.
- **`[pebble]` comment misattribution** — repointed to `..._Error`.
- **`[sand]` typed-path siblings** — added `TestDecoder_unsupportedFieldType` + `TestDecoder_untaggedFieldNoMatchingGroupSkipped`.
- **`[sand]` typed-error assertions** — switched to `errors.As`/`DecodeError` field checks (#142 contract).
- **`[dust]` cross-path godoc** — mirrored the note on `Decoder.One` (applies to `All`/`Iter`).

Full suite + `go vet` + `gofmt -s` + `golangci-lint` all clean. Per-thread replies above.

PR #159, opened 2026-06-30
feat: add ValidationError typed error for Validate (#151)

### LINKED ISSUE #151: Feature: typed *ValidationError with Missing []string for Validate
`Validate` reports missing required groups as an error string today, forcing callers to parse `err.Error()` to learn *which* groups are missing. Give it a typed error, exactly as #111 did for conversion failures with `DecodeError`:

```go
// proposed
type ValidationError struct {
    Missing []string // declared-but-absent required group names
}
func (e *ValidationError) Error() string
```

`Validate` returns an `errors.As`-able `*ValidationError` so callers can branch on the missing set programmatically instead of string-matching.

Same concern and precedent as #111. Message wording stays non-breaking per README §Stability (compare types, not strings).

Additive, non-breaking.

- Triaged from ROADMAP.md additive nice-to-haves via #144.
- Relates to: #111 (`DecodeError`, the typed-error precedent), #35 (`Validate`), the `regex:"name,required"` issue, #133.

## Implementation plan

**Approach.** Mirror `*DecodeError` (#111) exactly, scaled down. `Validate` already builds the ordered `missing []string` it needs; introduce an exported `type ValidationError struct{ Missing []string }` with an `Error() string` method whose text reproduces today's wording (`missing named groups: <comma-joined>`), and have `Validate` return `fmt.Errorf("regextra.Validate: %w", &ValidationError{Missing: missing})`. The `%w` wrap keeps the existing `regextra.Validate:` entrypoint prefix and the verbatim message while making the error `errors.As`-able. No `Unwrap` needed — unlike `DecodeError`, there is no underlying cause to surface (`Missing` *is* the payload).

**Steps.**
1. In `regextra.go`, add the `ValidationError` type + `Error()` method directly above `Validate`, with a godoc block patterned on `DecodeError`'s (errors.As example, note that `Missing` preserves the caller-passed order).
2. Change `Validate`'s terminal `fmt.Errorf` to wrap `&ValidationError{Missing: missing}` with the `regextra.Validate: %w` prefix; behavior and message are otherwise unchanged.
3. Update the `### Validate` README section: add an `errors.As(err, &ve)` snippet and note the typed error (the Stability §39 type-comparison guidance already covers it generically).
4. Add a CHANGELOG `### Added` entry citing #151 and the #111 precedent (additive, non-breaking).

**Files / surfaces touched.**
- `regextra.go` — new `ValidationError` type + `Error()`; `Validate` wraps it.
- `regextra_test.go` — new tests.
- `README.md` — `### Validate` section gains typed-error note + `errors.As` example.
- `CHANGELOG.md` — `### Added` entry.

**Verification.**
- New test: `errors.As(err, &ve)` recovers `*ValidationError` with `ve.Missing` equal (and in caller order) to the expected missing names; multi-missing case asserts order.
- Message non-breaking: assert the error string still has the `regextra.Validate:` prefix and contains `missing named groups: <names>` (loose match per README §Stability — don't pin exact text).
- `Validate` returning `nil` when nothing is missing still holds; `errors.As` against a `nil` error is false.
- `go test -race ./...`, `gofmt -s -l .`, `go vet ./...`, `golangci-lint run` clean; coverage stays above the 92% gate.

**Open questions / risks.**
- Public surface: this is a new exported type — confirm the field is `Missing []string` (matches the issue) and the type name is `ValidationError` (not `ValidateError`), consistent with `DecodeError`.
- Defensive copy of `Missing` is unnecessary: the slice is freshly allocated per `Validate` call and never retained internally, so handing it to the caller is safe.

**References.**
- #111 / PR #142 — `DecodeError`, the typed-error precedent this mirrors (type in `unmarshal.go`, README "Typed conversion failures", CHANGELOG Added entry).
- README §Stability — error wording is non-breaking; compare types/sentinels, not strings.
- Current `Validate` impl: `regextra.go` ~L614–645.

Jecoms:
## Summary
- Add an exported `type ValidationError struct{ Missing []string }` with an `Error()` method, placed next to `Validate` in `regextra.go`.
- `Validate` now returns `fmt.Errorf("regextra.Validate: %w", &ValidationError{Missing: missing})`, so callers can recover the typed error with `errors.As` and branch on the missing set instead of parsing `err.Error()`. The prefix and message wording are unchanged — additive, non-breaking.
- Mirrors the `DecodeError` precedent (#111). No `Unwrap`: `Missing` is the payload, there's no underlying cause.
- README §Validate gains a typed-error note + `errors.As` example; CHANGELOG `### Added` entry added.

## Test plan
- [x] New `TestValidateValidationError`: `errors.As` recovers `*ValidationError` with `Missing` in request order (single + multi-missing), and `errors.As` against a nil error is false. Message asserted to keep the `regextra.Validate:` prefix and `missing named groups: …` text (loose match per §Stability).
- [x] `go test -race ./...` passes; coverage 98.1% (gate 92%).
- [x] `gofmt -s -l .` clean, `go vet ./...` clean, `golangci-lint run` 0 issues, `go build ./...` clean.

Closes #151

## Summary by CodeRabbit

* **New Features**
  * Validation errors now expose structured missing-group details, making it easier to inspect which required names were not found.
  * Error handling now supports checking for the validation error type directly, instead of relying only on message text.

* **Documentation**
  * Updated the changelog and usage docs with examples showing how to detect and read missing group names from validation errors.


regextra.go
```diff
@@ -611,13 +611,40 @@ func replaceNamed(re *regexp.Regexp, target string, replFor func(name, matched s
...
+type ValidationError struct {
+	// Missing holds the declared-but-absent required group names, in the order
+	// they were passed to Validate.
+	Missing []string
+}
+
+// Error implements the error interface. Validate prepends its own
+// `regextra.Validate:` prefix when wrapping.
+func (e *ValidationError) Error() string {
+	return fmt.Sprintf("missing named groups: %s", strings.Join(e.Missing, ", "))
```
Jecoms, line 643: `[dust]` Because `ValidationError` is exported, a directly-constructed `&ValidationError{}` (empty/nil `Missing`) renders `Error()` as `"missing named groups: "` with a dangling trailing space. `Validate` itself never produces this (it only wraps when `len(missing) != 0`), and `*DecodeError` carries the same direct-construction exposure, so this is awareness-only — flagging in case a guard or `len`-based message is wanted. (CONTRIBUTING §Documentation; mirrors the #111 `DecodeError` precedent.)

Jecoms (commented):
### Code review

**1 finding** — 1 dust. Inline comment below.

Clean, tightly-scoped addition that mirrors the `*DecodeError` precedent (#111) well: `Validate`'s logic is unchanged — the `missing` slice is still built fresh per call (no aliasing) and only the wrapping changes from a flat `fmt.Errorf` to `%w` around `&ValidationError{Missing: missing}`. The `regextra.Validate:` prefix and `missing named groups: …` text are preserved, so the change is additive and non-breaking per the type-not-string contract in [§Stability](https://github.com/Jecoms/regextra/blob/3997ffe988b3131bfabf7033afb667ef816c1cb2/README.md#L18-L40). Godoc covers every exported symbol (type, `Error()`, `Missing` field); the new test lands in the correct `regextra_test.go` sibling per CONTRIBUTING's one-test-file-per-source rule and asserts only the exported contract (`errors.As`, request-order `Missing`, prefix preserved, nil-error negative case). No runnable `ExampleValidationError` is consistent with `DecodeError`, which also ships only an inline godoc example. `Validate` is an init-time/cold path, so no benchmark gate applies.

regextra.go
```diff
@@ -611,13 +611,40 @@ func replaceNamed(re *regexp.Regexp, target string, replFor func(name, matched s
...
+//
+//	var ve *regextra.ValidationError
+//	if errors.As(err, &ve) {
+//	    log.Printf("pattern is missing groups: %v", ve.Missing)
+//	}
+//
+// Missing lists the absent names in the order they were passed to Validate.
+// Unlike [DecodeError] there is no underlying cause to unwrap — Missing is the
+// payload.
+type ValidationError struct {
```
Jecoms: [rock] This custom error is very generic but being used in a very specific way. Please update this custom error name to be specific to the error scope.

regextra.go
Jecoms: Good catch — addressed in 24aa12e (follow-up). Rather than diverge from the `*DecodeError` precedent, I guarded **both** sites so the sibling typed errors stay consistent:

- `ValidationError.Error()` now returns `no missing named groups` when `Missing` is empty/nil (instead of the dangling `missing named groups: `).
- `DecodeError.Error()` now returns `no decode error` when `Err` is nil (instead of `field : <nil>`).

Both guards only fire on direct construction of an empty value — `Validate`/`Unmarshal` never build an empty payload, so every real path renders exactly as before (non-breaking per §Stability). Added `Error()` empty-case tests in `regextra_test.go` and `unmarshal_test.go` (coverage 98.1%), and noted the empty-construction behavior on both CHANGELOG `[Unreleased]` Added bullets.

regextra.go
Jecoms: Renamed in ad6d969 (follow-up): `ValidationError` → `MissingNamedGroupsError`. The new identifier matches what the error actually reports — the required named capture groups `Validate` could not find declared — and the rendered `missing named groups: …` message, paralleling the `DecodeError` `<Scope>Error` convention.

Applied everywhere: the type def + `Error()` receiver + the `&MissingNamedGroupsError{Missing: …}` construction in `Validate` and its godoc (regextra.go), the test refs and `TestValidateMissingNamedGroupsError` (regextra_test.go), the README example, and the CHANGELOG `[Unreleased]` bullet. The type is new/unreleased, so this is non-breaking; the `Missing` field, the message text, the `regextra.Validate:` prefix, and the empty-case `no missing named groups` guard are all unchanged. Checks green from the worktree (gofmt -s -l, go vet, go build, go test -race, golangci-lint 0 issues; coverage 98.1%).

PR #160, opened 2026-06-30
feat: tag-derived required-group validation (regex:",required")

### LINKED ISSUE #148: Feature: tag-derived required-group validation (regex:"name,required")
`Unmarshal`/`Decoder` callers who need a group to be present must today make a separate `Validate` call. Let a struct field declare that requirement inline on its tag:

```go
type T struct {
    Name string `regex:"name,required"`
}
```

A missing/non-participating `required` group makes the decode fail (surfacing through the typed validation error — see the `*ValidationError` issue), removing the need for a separate `Validate` pass in the common case.

This is the next-minor headline feature. The forward-compat groundwork is already pinned as v1 contract: the README §Tag grammar and the package doc (`regextra.go`) document the lone-token "silently-ignored-today, promote-to-flag-later" plan, so `regex:"name,required"` recognizing the `required` token is additive and non-breaking.

Additive, non-breaking.

- Triaged from ROADMAP.md "Next minor" via #144.
- Relates to: #133 (deferred-work bundle), the typed `*ValidationError` issue.

## Implementation plan

**Approach.** Promote the reserved `required` lone token to a recognized flag, carry it through the shared decode plan, and enforce it at the single point in `runDecodePlan` where a group yields no value. `required` fires when the group did not participate, matched an empty span, or is undeclared **and** no `default=` is set (a `default` produces a value and satisfies the requirement). The one enforcement hook covers all five entrypoints (`Unmarshal`, `UnmarshalAll`, `Decoder.One/All/Iter`), each inheriting its existing `regextra.<Entrypoint>:` error wrapping.

**Steps.**
1. `parseFieldTag` (`unmarshal.go:265-294`): in the lone-token branch recognize `required`; return a new `required bool`. Update both call sites (`buildDecodePlan`, `resolveGroupName`).
2. `fieldDecoder` (`decoder.go:46-60`): add `required bool`. In `buildDecodePlan` set it, and change the no-op skip (`decoder.go:193`) to keep required fields: `if len(groupIdxs) == 0 && !hasDefault && !required { continue }`.
3. `runDecodePlan` (`decoder.go:382-385`): replace `if !ok { continue }` with — if `!ok && fd.required`, return a typed `*ValidationError{Field, Group}` (group via `resolveGroupName`); else `continue`.
4. Add a `ValidationError` type beside `DecodeError` (`unmarshal.go:65-96`), mirroring its typed-error contract (`errors.As`-recoverable; optional `ErrRequiredGroupMissing` sentinel + `Unwrap` for `errors.Is`). The repo contract is typed errors, not message strings.

**Files / surfaces touched.**
- `unmarshal.go` — `parseFieldTag` flag; new `ValidationError`.
- `decoder.go` — `fieldDecoder.required`, plan retention, enforcement hook.
- `regextra.go` (§Tag grammar pkg doc, ~152-174) & `README.md` (options table ~249-252, forward-compat bullet ~266) — promote `required`; swap the lone-token examples to a still-reserved token (e.g. `foo`).
- `CHANGELOG.md` — "Added" entry under next minor.

**Verification.** `go test ./...`. Update `TestUnmarshal_loneTokenIsIgnored` / `TestDecoder_tagForwardCompatRules` to use a still-reserved token. Add tests (mirroring `TestUnmarshalDecodeError` / `TestDecoderDecodeError`): required absent/non-participating → `errors.As` a `*ValidationError` under the right entrypoint prefix; `required,default=x` still succeeds; required satisfied when the group participates; strict `Compile` still rejects `required` on an undeclared group with no default.

**Open questions / risks.**
- Empty-but-participating semantics: plan keys on the existing `ok` signal (empty span fails `required`). If empty-but-present should satisfy `required`, key on `found` instead — confirm with maintainer.
- The issue names `*ValidationError`; this introduces it. If a sibling issue owns that type, coordinate so the definition isn't duplicated.

**References.**
- Shared decode plan: #143. `DecodeError`/error-prefix contract: #142, #111. Forward-compat token reservation: README §Tag grammar + `regextra.go` pkg doc.

Jecoms:
## Summary

Implements #148: promote the reserved `required` lone-token tag flag to a recognized flag so a struct field can declare its capture group mandatory inline (`regex:"name,required"`), removing the need for a separate `Validate` pass in the common case.

- **`parseFieldTag`** now recognizes the `required` lone token (the first reserved flag slot the forward-compat policy set aside) and returns a `required bool`. Any *other* lone token stays silently ignored.
- **`fieldDecoder.required`** carries the flag through the shared decode plan; `buildDecodePlan` retains a required field even when it maps to no group/default so it can be enforced.
- **`runDecodePlan`** enforces at the single no-value point: when a required field yields no value, it returns a new typed `*RequiredGroupError{Field, Group}`, wrapped under each entrypoint's `regextra.<Entrypoint>:` prefix. One hook covers all five entrypoints (`Unmarshal`, `UnmarshalAll`, `Decoder.One`/`All`/`Iter`).
- **`RequiredGroupError`** is a new `errors.As`-able typed error, mirroring `DecodeError`'s style. It is the per-match presence check, distinct from `*DecodeError` (a participating value that failed conversion) and `*MissingNamedGroupsError` (the static `Validate` check). Per coordinator decision, it is a new purpose-specific type — the already-shipped `MissingNamedGroupsError` (#151) stays scoped to `Validate` and is not reused/renamed.
- Docs updated: package doc Tag-grammar section, README Tag-options table + forward-compat bullet + a "Required groups" typed-error subsection, and a CHANGELOG `Added` entry.

### Semantics

- A `default=` satisfies `required` (it always yields a value).
- Presence keys on the existing `resolveGroupValue` "empty span = data absence" contract, so a **participating-but-empty span also fails `required`** (consistent with how the skip-or-default path already treats empty spans). Covered by a dedicated test.
- No match is data absence, not a failure — `Unmarshal` returns nil without running the plan, so `required` does not fire on a non-match.
- Strict `Compile` still rejects `required` on an undeclared group with no default (unchanged undeclared-group check).

## Test plan

- [x] `go build ./...`, `go vet ./...`, `gofmt -s -l .` clean
- [x] `go test -race ./...` passes; coverage 99.2% (> 92% gate)
- [x] Hot-path benchmarks (`BenchmarkUnmarshal*` / `BenchmarkDecoder*`) hold — same allocation profile as baseline; the added field + rare-path guard add no measurable cost
- [x] New tests: `TestUnmarshalRequired` (Unmarshal/UnmarshalAll — non-participating, empty-span, undeclared, satisfied, default, no-match, match-prefix) and `TestDecoderRequired` (One/All/Iter prefixes, satisfied, default, strict-Compile rejection); updated the now-stale forward-compat tests that used `required` as an inert lone token to use `foo`

Closes #148

## Summary by CodeRabbit

* **New Features**
  * Added support for marking regex capture groups as `required` in struct tags.
  * Introduced clearer decode errors when a required group is missing, doesn’t participate, or matches an empty span without a default value.

* **Documentation**
  * Updated the README and changelog with examples and guidance for the new tag option.
  * Clarified how required-group errors differ from other decoding and validation errors.


unmarshal.go
```diff
@@ -101,6 +101,45 @@ func (e *DecodeError) Error() string {
...
+// Error implements the error interface. The calling entrypoint prepends its own
+// `regextra.<Entrypoint>:` prefix when wrapping. When Field is empty (only
+// reachable by constructing the value directly — the decode path always sets the
+// field name) it reports "no required group error" rather than a message about a
+// nameless field.
+func (e *RequiredGroupError) Error() string {
+	if e.Field == "" {
+		return "no required group error"
+	}
+	return fmt.Sprintf("field %s: required group %q did not match", e.Field, e.Group)
```
Jecoms: `[dust]` The message reads `required group %q did not match`, but a required field can also fail when its group *did* match an empty span (the documented participating-but-empty-span path that `runDecodePlan` keys on `resolveGroupValue`'s `ok`). "did not match" is imprecise for that case — something like "produced no value" covers both modes. (source: this method's own contract + the empty-span semantics in the PR description / `runDecodePlan` comment)

unmarshal.go
```diff
@@ -101,6 +101,45 @@ func (e *DecodeError) Error() string {
...
 func (e *DecodeError) Unwrap() error { return e.Err }
 
+// RequiredGroupError reports that a field marked `regex:",required"` produced no
+// value for a match: its capture group did not participate, or matched an empty
+// span, and no `default=` supplied a substitute. It is returned (wrapped with the
+// calling entrypoint's prefix) by [Unmarshal], [UnmarshalAll], [Decoder.One],
+// [Decoder.All], and [Decoder.Iter]. Recover it with [errors.As] to branch on
+// the missing field without parsing message text:
+//
+//	var re *regextra.RequiredGroupError
```
Jecoms: `[dust]` The godoc example binds the error to `re`, the name used package-wide for the `*regexp.Regexp`; the README "Required groups" example uses `rge` and `DecodeError`'s godoc uses `de`. A non-`re` name would avoid shadowing the regexp idiom and match the sibling examples. (source: README Required-groups example / `DecodeError` godoc style)

Jecoms (commented):
### Code review

**2 finding(s)** — 2 dust. Inline comments below.

Well-scoped feature: the `required` flag threads cleanly through `parseFieldTag` -> `fieldDecoder` -> `buildDecodePlan` -> `runDecodePlan`, the single enforcement point keying on `resolveGroupValue`'s `ok` (so empty-span = absence) is consistent with the existing skip/default contract, strict `Compile` still rejects undeclared-group-no-default, and `default=` correctly short-circuits. Tests are routed to the right sibling files per CONTRIBUTING.md, drive the public API, and cover the matrix (non-participating, empty span, undeclared, satisfied, default, no-match, all five entrypoints). Docs (package doc, README table + forward-compat bullet, CHANGELOG) are all updated. Nothing blocking — only two cosmetic doc nits below.

unmarshal.go
Jecoms: Addressed in 256c01f0 — reworded the message to `required group %q produced no value`, which covers both the non-participating and participating-but-empty-span failure modes (and matches the type's own godoc wording).

unmarshal.go
Jecoms: Addressed in 256c01f0 — renamed the godoc example error variable from `re` to `rge` to avoid shadowing the package-wide `*regexp.Regexp` idiom and align with the README "Required groups" example and the sibling `DecodeError`/test conventions.

Jecoms:
Added a hot-path benchmark for the `required` flag — and it caught a regression, now fixed. Two follow-up commits:

- **`perf(unmarshal): allocate tag options map lazily`** — `parseFieldTag` was allocating an empty `map[string]string` for any field whose tag had a second comma-part, including a lone-flag-only tag like `name,required` (the flag never populates the map). Since `Unmarshal` reparses tags every call, that was **+1 alloc per required field per call** on the decode hot path. The map is now allocated lazily — only a `key=value` pair populates it — so a lone-flag field keeps `opts` nil, matching the no-options case. Behavior unchanged (consumers already treat nil opts as "no options").
- **`test(unmarshal): benchmark the required flag on the decode hot path`** — `BenchmarkUnmarshal/requiredFields`, a `benchSimple` clone with all three fields `required`, sharing the same pattern/input so a `benchstat` diff against `simpleThreeField` isolates the flag's cost.

`benchstat`-style, `-count=6` (Apple M4):

| case | ns/op | allocs/op |
|---|---|---|
| `simpleThreeField` (baseline) | ~540 | 11 |
| `requiredFields` — before fix | ~645 | 14 |
| `requiredFields` — after fix | ~595 | **11** |

Hot-path allocation-neutrality restored (per CONTRIBUTING §131). Gates: gofmt/build/vet clean, `go test -race` green, golangci-lint 0 issues, coverage 99.2%. The failure path (`*RequiredGroupError` construction) is intentionally not benchmarked — it's cold error construction, covered by tests.

PR #164, opened 2026-07-01
perf: trim off-hot-path allocations in AllNamedGroups

### LINKED ISSUE #154: Perf: micro-optimize off-hot-path allocations (NamedGroups and Find* helpers)
The unmarshal hot path has its own perf tracking; this issue is for the off-hot-path allocations the roadmap backlog called out — e.g. `NamedGroups` and the other `Find*`/`NamedGroups*` helpers building a fresh map (and `SubexpNames` slice) per call.

Scope: micro-optimize allocations in the non-`Unmarshal` helpers without changing behavior or signatures. Any change must be backed by benchmark deltas per CONTRIBUTING (benchmark validation is required for hot-path/feature changes).

Lower priority — these aren't on the typed-unmarshal hot path. Pull forward if a profile shows real adopter pain.

Non-breaking (internal only).

- Triaged from ROADMAP.md "Backlog" via #144.
- Relates to: #133.

## Implementation plan

**Approach.** Signature-preserving, behavior-preserving allocation trims to the non-`Unmarshal` helpers, gated on the existing `regextra_bench_test.go` benchmarks. The single clear win is `AllNamedGroups` (regextra.go:462), which still calls `re.FindStringSubmatch` — allocating a `[]string` of every group per call — while its siblings `FindNamed`/`FindAllNamed` already switched to `FindStringSubmatchIndex` and slice `target` directly. Switch it to the index form and presize its result map/slices. `re.SubexpNames()` returns the regexp's cached slice (no copy), so the "SubexpNames slice per call" cost noted in the issue is a couple of pointer derefs, not an allocation — hoist the double call in `Validate` for tidiness only, don't over-claim a win there. Land only the deltas that benchmarks actually show; drop any that don't move `allocs/op`.

**Steps.**
1. `AllNamedGroups`: replace `FindStringSubmatch` with `FindStringSubmatchIndex`; build values by slicing `target[start:end]` guarding `start >= 0` for non-participating groups (mirror `fillNamedGroupValues`). Presize `result` with a name count and give each slice cap ~1.
2. Micro-tidy (optional, only if benched neutral-or-better): hoist the two `re.SubexpNames()` calls in `Validate` (regextra.go:629-630) into one local; consider a small-slice fast path in `subexpIndexes` (decoder.go:211) only if a bench shows the per-call `[]int` append allocation matters — otherwise leave it.
3. Run `go test -bench=. -benchmem` before/after; keep only changes that reduce `allocs/op` or `B/op` without behavior change.

**Files / surfaces touched.**
- `regextra.go` — `AllNamedGroups` (index-form rewrite + presize); optional `Validate` hoist.
- `decoder.go` — optional `subexpIndexes` fast path (only if benched worthwhile).
- No public signatures change; internal-only.

**Verification.**
- Existing `regextra_test.go` cases for `AllNamedGroups` (distinct + duplicate group names, no-match empty-map) stay green under `go test -race ./...`; coverage stays >92%.
- `go test -bench=BenchmarkAllNamedGroups -benchmem` (and the other `Benchmark{Find,NamedGroups}*`) shows equal-or-lower `allocs/op`/`B/op`; record the delta in the PR body per CONTRIBUTING step 5.

**Open questions / risks.**
- `AllNamedGroups` returns a `map[string][]string` and duplicate names append per occurrence — the map + per-name slices are inherent to the return type and cannot be eliminated; the win is only the discarded full-submatch `[]string`. Confirm the bench delta is real before claiming it.
- Per-match map allocation in `NamedGroupsPerMatch`/`NamedGroupsPerMatchSeq` is inherent to their return shapes (slice/stream of maps) — out of scope; do not restructure.
- If no change produces a measurable delta, close as "no worthwhile micro-opt" rather than landing churn (p3, non-breaking).

**References.**
- CONTRIBUTING.md — benchmark validation (PR step 5); one `_bench_test.go` per source file.
- Related: #133; triaged from ROADMAP backlog via #144.

Jecoms:
## Summary
- Switch `AllNamedGroups` from `FindStringSubmatch` to `FindStringSubmatchIndex`, slicing each participating group out of `target` directly — the same index-form pattern its `FindNamed`/`FindAllNamed` siblings already use. This drops the full-submatch `[]string` allocated per call. Behavior is preserved, including contributing `""` for declared-but-non-participating groups.
- Drop the map size hint idea from the plan: `len(names)` counts name *slots*, not distinct keys, so presizing over-allocated the map when a pattern reuses one name (it regressed `manyDuplicates`). Removed to keep every case neutral-or-better.
- Tidy: hoist the double `re.SubexpNames()` call in `Validate` into one local (no alloc change).
- `decoder.go`/`subexpIndexes` fast path from the plan was left out — it was gated on "only if a bench shows it matters", and no benched win materialized, so no churn.

## Benchmarks
`go test -run=^$ -bench='BenchmarkAllNamedGroups|BenchmarkValidate' -benchmem -benchtime=100ms -count=6` (Apple Silicon, arm64). Per CONTRIBUTING benchmark validation:

| case | before | after |
|---|---|---|
| AllNamedGroups/distinctGroups | 548 B/op, 6 allocs/op | 482 B/op, 5 allocs/op |
| AllNamedGroups/duplicateGroupName | 644 B/op, 7 allocs/op | 579 B/op, 6 allocs/op |
| AllNamedGroups/manyDuplicates | 2128 B/op, 10 allocs/op | 1772 B/op, 9 allocs/op |
| AllNamedGroups/noMatch | 48 B/op, 1 allocs/op | 48 B/op, 1 allocs/op (neutral) |
| Validate/* | unchanged | unchanged |

Each contributing case drops one alloc and B/op; no regressions.

## Test plan
- [x] `go test -race ./...` green
- [x] `go vet ./...` clean, `gofmt -l` clean
- [x] `go test -cover .` = 98.1% (>92% floor); existing `AllNamedGroups` cases (distinct, duplicate, no-match) cover behavior preservation
- [x] Before/after benchmarks recorded above

Fixes #154

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved named-group extraction so empty but present groups are now preserved correctly in results.
  * Ensured only the first match is used consistently when collecting group values.

* **Refactor**
  * Streamlined validation logic for better internal efficiency without changing behavior.

### DIFF
--- a/regextra.go
+++ b/regextra.go
@@ -437,8 +437,8 @@ func fillNamedGroupValues(dst map[string]string, names []string, target string, 
 // Groups that appear once still get a one-element slice.
 //
 // The leading "All" refers to all named groups in one match — not to all
-// matches across the target. Internally the function calls FindStringSubmatch,
-// so only the first match contributes values. To collect every value of a
+// matches across the target. Internally the function inspects only the first
+// match, so only the first match contributes values. To collect every value of a
 // single named group across every match in the target, use [FindAllNamed]. To
 // collect every named group across every match as one map per match, use
 // [NamedGroupsPerMatch] (or [NamedGroupsPerMatchSeq] for the lazy form); the
@@ -460,17 +460,30 @@ func fillNamedGroupValues(dst map[string]string, names []string, target string, 
 //	allGroups := regextra.AllNamedGroups(re, "Alice 30")
 //	// allGroups = map[string][]string{"name": []string{"Alice"}, "age": []string{"30"}}
 func AllNamedGroups(re *regexp.Regexp, target string) map[string][]string {
+	names := re.SubexpNames()
+	// No map size hint: len(names) counts slots, not distinct names, so it
+	// over-allocates when a pattern reuses one name across many occurrences.
 	result := make(map[string][]string)
 
-	matches := re.FindStringSubmatch(target)
-	if matches == nil {
+	// Index form returns only the match offsets ([]int); we slice each
+	// participating group out of target directly. FindStringSubmatch would
+	// additionally allocate a []string of every group just to read it once.
+	loc := re.FindStringSubmatchIndex(target)
+	if loc == nil {
 		return result
 	}
 
-	for i, name := range re.SubexpNames() {
-		if i != 0 && name != "" {
-			result[name] = append(result[name], matches[i])
+	for i, name := range names {
+		if i == 0 || name == "" {
+			continue
 		}
+		// A non-participating group has start < 0; preserve
+		// FindStringSubmatch's behavior of contributing "" for it.
+		value := ""
+		if start := loc[2*i]; start >= 0 {
+			value = target[start:loc[2*i+1]]
+		}
+		result[name] = append(result[name], value)
 	}
 
 	return result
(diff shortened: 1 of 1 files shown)


Jecoms (commented):
### Code review

No findings. Checked convention-doc compliance, the diff, git history, prior
PR feedback, and code comments.

PR #166, opened 2026-07-01
feat: sentinel errors for Compile/MustCompile failure categories

### LINKED ISSUE #152: Feature: sentinel errors for Compile/MustCompile failure categories
`Compile`/`MustCompile` can fail for distinct reasons (invalid regexp, struct/tag-shape problems, etc.), but callers can only inspect the error string. Expose sentinel errors so failure categories are `errors.Is`-checkable:

```go
// proposed, illustrative
var (
    ErrInvalidPattern = errors.New("regextra: invalid pattern")
    ErrInvalidStruct  = errors.New("regextra: invalid struct")
)
```

Wrap the underlying cause so `errors.Is(err, ErrInvalidPattern)` works while the message keeps its detail. Implementer should enumerate the actual failure categories `compileDecoder`/`buildDecodePlan` produce before fixing the sentinel set.

Consistent with the typed-error contract (#111) — branch on sentinels/types, not strings.

Additive, non-breaking.

- Triaged from ROADMAP.md additive nice-to-haves via #144.
- Relates to: #111, #31 (`Compile`), #133.

## Implementation plan

_Added 2026-07-01 via /gh-analyze-issue. Review before implementing._

**Approach.** The Compile path (`Compile` → `regexp.Compile` + `compileDecoder` → `buildDecodePlan(strict=true)`) has 5 error-return sites collapsing into 2 caller-facing categories: bad regexp vs. bad destination struct. Introduce two package sentinels and wrap each site with `fmt.Errorf` + `%w`, leading with the sentinel so the existing single-line `regextra.Compile: …` message detail is preserved and `errors.Is` works. `MustCompile` inherits this (it panics with the same error). `buildDecodePlan(strict=false)` on the Unmarshal path never errors, so sentinels stay Compile-only.

**Enumerated failure categories** (verified against `decoder.go`):
- `ErrInvalidPattern` → P1: `regexp.Compile` failure (invalid regexp).
- `ErrInvalidStruct` → P2: `T` not a struct; P3: field references undeclared group; P4: `default=` does not convert; P5: `layout=` on a non-`time.Time` field.

**Steps.**
1. Add `ErrInvalidPattern` / `ErrInvalidStruct` `var`s next to `ErrNoMatch` in `decoder.go`, doc-commented, keeping the bare-`regextra:` sentinel prefix convention.
2. Wrap the 5 sites with `%w`; for P1 and P4 (which already wrap a cause) use two `%w` verbs (Go 1.24) so both sentinel and underlying cause stay reachable.
3. Update the `Compile` doc comment to name the sentinels callers can `errors.Is` against.
4. Add `errors.Is` tests per category + a `MustCompile` recover test; assert Unmarshal on a bad-tag struct does NOT surface `ErrInvalidStruct`.
5. Update `CHANGELOG.md` (Unreleased → Added) and `README.md` (§Stability + error section) sentinel list.

**Files / surfaces touched.**
- `decoder.go` — 2 sentinels, 5 wrap sites, `Compile` doc.
- `decoder_test.go` — `errors.Is`/recover tests (existing `strings.Contains` tests stay green; substrings preserved).
- `CHANGELOG.md`, `README.md` — public-contract docs.

**Verification.** `go test ./...` and `go vet ./...` (multiple `%w` valid under Go 1.24). New tests assert `errors.Is` per category, preserved underlying cause for P1/P4, and that sentinels never leak from the Unmarshal path.

**Open questions / risks.**
- Sentinel granularity: plan uses **2 sentinels** (matches the issue's examples and is smallest-correct). A **3-sentinel** split (`ErrInvalidPattern` / `ErrInvalidType` for P2 / `ErrInvalidTag` for P3–P5) is a later additive option — confirm 2 is acceptable before implementing.
- Cosmetic double prefix: leading with a `regextra:`-prefixed sentinel yields `regextra.Compile: regextra: invalid pattern: …`. Non-breaking (don't string-match) and consistent with `ErrNoMatch`; keep unless the reviewer objects.
- Do not use `errors.Join` (breaks the standardized single-line prefix format from #111).

**References.**
- #111 typed-error contract (`DecodeError`, standardized `regextra.<Entrypoint>:` prefixes; bare `regextra:` reserved for sentinels).
- `README.md` §Stability: compare against exported sentinels with `errors.Is`, not `err.Error()`.

Jecoms:
## Summary

- Add two package sentinels, `ErrInvalidPattern` and `ErrInvalidStruct`, and wrap `Compile`'s five validation error sites with them via `fmt.Errorf(... %w ...)`, so callers can branch on the failure kind with `errors.Is` instead of parsing `err.Error()`.
  - `ErrInvalidPattern` → invalid regular expression.
  - `ErrInvalidStruct` → the four destination-shape problems: `T` is not a struct, a field references an undeclared group, a `default=` value does not convert, and `layout=` on a non-`time.Time` field.
- The pattern-compile and bad-`default=` sites use a two-`%w` wrap (Go 1.24) so both the sentinel and the underlying cause stay reachable. Every message keeps its existing detail, preserving #111's single-line `regextra.<Entrypoint>:` prefix contract; the sentinels themselves carry the bare `regextra:` prefix reserved for sentinels (like `ErrNoMatch`).
- `MustCompile` inherits this — it panics with the same wrapped error. Sentinels are Compile-only: the lenient `Unmarshal` / `UnmarshalAll` path tolerates the same shapes and never surfaces them.
- Docs updated: `Compile` godoc, `README.md` (§Stability sentinel list + error section with an `errors.Is` branch example), and `CHANGELOG.md` (Unreleased → Added).

Additive, Compile-only, non-breaking. The per-call decode/match hot path is untouched.

## Test plan

- [x] `go vet ./...` clean (multiple `%w` valid under Go 1.24).
- [x] `go test ./...` passes; existing `strings.Contains` Compile tests stay green (substrings preserved).
- [x] New tests: `errors.Is` per category (pattern vs. struct, and that they don't cross-match), preserved underlying cause for the pattern-compile and bad-`default=` cases, a `MustCompile` recover-and-`errors.Is` test, and an assertion that the `Unmarshal` path never leaks a Compile sentinel.
- [x] `BenchmarkCompile*` / `BenchmarkDecoder*` (`-benchmem -count=3`) hold — decode hot path allocation-neutral (change is error-path only).

Fixes #152


decoder.go
```diff
@@ -70,10 +85,15 @@ type fieldDecoder struct {
...
+//
+// Failures are categorized by wrapped sentinel: the first cause above wraps
+// [ErrInvalidPattern] and the remaining four wrap [ErrInvalidStruct], so
+// callers can branch on the failure kind with errors.Is instead of parsing the
+// message. [MustCompile] panics with the same wrapped error.
 func Compile[T any](pattern string) (*Decoder[T], error) {
 	re, err := regexp.Compile(pattern)
 	if err != nil {
-		return nil, fmt.Errorf("regextra.Compile: invalid pattern: %w", err)
+		return nil, fmt.Errorf("regextra.Compile: %w: %w", ErrInvalidPattern, err)
```
Jecoms: `[pebble]` Wrapping a sentinel that itself carries the bare `regextra:` prefix under the `regextra.Compile:` entrypoint prefix renders a doubled prefix: `regextra.Compile: regextra: invalid pattern: missing closing ]`. Unlike `ErrNoMatch` — which is returned bare and never entrypoint-wrapped (decoder.go:288), so it never doubles — these Compile sentinels are *always* wrapped, so the `regextra:` on them is redundant in every rendered message. Same at lines 117, 184, 195, 206. Two documented conventions collide here (the entrypoint-prefix scheme and the bare-`regextra:`-for-sentinels scheme, both described in README.md's Stability/error section); pick one for the wrap. Cleanest options: wrap without the entrypoint prefix (`fmt.Errorf("%w: %w", ErrInvalidPattern, err)`) to match how `ErrNoMatch` is surfaced, or drop `regextra:` from the sentinel messages since they're never printed standalone. Purely cosmetic — the `errors.Is` contract is unaffected. (Source: README.md error-prefix section + `ErrNoMatch` precedent at decoder.go:288.)

Jecoms (commented):
### Code review

**2 finding(s)** — 1 pebble, 1 dust. See the threaded comments below (and the note not tied to a changed file).

Solid, well-scoped additive change: two `errors.Is`-checkable sentinels wrapped at the five `Compile` validation sites, with the underlying cause preserved via two-`%w` where one exists, thorough per-category tests (including the `MustCompile` recover path and the Unmarshal-doesn't-leak assertion), and matching README/CHANGELOG/godoc. The typed-error contract (compare sentinels, don't parse strings) is honored, so the substring `strings.Contains` tests stay green without pinning wording. One cosmetic prefix-doubling point below.

1. `[dust]` The PR currently reports `mergeable: CONFLICTING` against `main` — likely a `CHANGELOG.md`/`README.md` overlap with a recently-landed PR. A rebase will be needed before merge. (PR state via `gh pr view 166`.)

decoder.go
Jecoms: Addressed in 837d881 (option a). Dropped the `regextra.Compile:` entrypoint prefix from all five sentinel wrap sites, so they now surface bare (`fmt.Errorf("%w: …", ErrInvalidPattern/ErrInvalidStruct, …)`) and render single-prefixed, e.g. `regextra: invalid pattern: missing closing ]` — matching how `ErrNoMatch` is returned. Chose (a) over dropping `regextra:` from the sentinel messages because the godoc (decoder.go) and CHANGELOG already document these sentinels as carrying the bare `regextra:` prefix like `ErrNoMatch`, and README §Stability reserves that prefix for package-level sentinels — (a) keeps all of that accurate. `errors.Is` contract unchanged; the substring tests stay green (none pinned the entrypoint prefix).

Jecoms:
Addressed the review in follow-up commit `837d881`:

- **[pebble] double `regextra:` prefix** — dropped the `regextra.Compile:` entrypoint prefix from all five sentinel wrap sites so they render single-prefixed like `ErrNoMatch` (option a). See the inline reply on that thread.
- **[dust] CONFLICTING with main** — rebased onto latest `main` (now includes #164 plus the `required`-group / `MissingNamedGroupsError` work). Resolved the `CHANGELOG.md` and `README.md` overlap by keeping the new `Added` entries and slotting the `ErrInvalidPattern`/`ErrInvalidStruct` entry among them. PR is now `MERGEABLE`.

`go vet ./...` and `go test ./...` both green.

Issue #149, opened 2026-06-30
Feature: Encoder[T] typed round-trip via template syntax


Jecoms:
`Decoder[T]` turns a matched string into a typed struct. The symmetric counterpart would build a string from a typed struct via template syntax:

```go
// proposed, speculative
type Encoder[T any] struct{ /* ... */ }
func NewEncoder[T any](template string) (*Encoder[T], error)
func (e *Encoder[T]) Encode(v T) (string, error)
```

i.e. typed round-trip: `Encode` then `Unmarshal`/`Decode` returns the original struct.

Speculative — roadmap marked this "possibly". Ships only on concrete adopter demand; filing to track the idea, not to schedule it. Comment with a use case to pull it forward.

Additive, non-breaking.

- Triaged from ROADMAP.md "Next minor (possibly)" via #144.
- Relates to: #133 (deferred-work bundle).

## Implementation plan

_Added 2026-07-01 via /gh-analyze-issue. Review before implementing. Still gated on concrete adopter demand — this plans the **how**, it does not schedule the work._

**Approach.** Add `Encoder[T]` as the inverse of `Decoder[T]`, reusing the existing tag/field-mapping machinery (`parseFieldTag`, `matchGroupName`) rather than adding new mapping rules. The crux is that the library has **no template syntax today** — patterns are raw `regexp` strings with `(?P<name>...)` groups, which are not generically reversible. So `NewEncoder[T]` compiles a **new, reversible mini-template**: literal text kept verbatim plus named placeholders that resolve to fields exactly as group names do. Parse once at construction into an ordered segment list; `Encode` walks segments, converts each field via a new `encodeFieldValue` (the inverse of `setFieldValue`), and concatenates with `strings.Builder` to stay allocation-light.

**Scope for v1.** Literal text + **required** named placeholders separated by literals (e.g. `"{name} is {age}"`). Defer regex optional groups and `|` alternation (no unambiguous encode form). Round-trip guarantee holds when each encoded value re-matches its placeholder's sub-pattern; document that adjacent placeholders / delimiter-collision values are out of scope for v1.

**Steps.**
1. Design the template mini-language: pick a delimiter (`{name}` recommended), define escaping for literal braces, and the placeholder→field resolution (reuse `parseFieldTag` + `matchGroupName` semantics, honoring `regex:"..."` tags and `-` exclusion).
[...]

PR #168, opened 2026-07-01
feat(encoder): derive Encoder[T] from the decoder's compiled pattern

Jecoms:
## Summary

Adds `Encoder[T]`, the typed inverse of `Decoder[T]` (issue #149), **derived from the decoder's own compiled pattern** rather than a hand-written template. Write the pattern once and get the inverse for free:

```go
dec := regextra.MustCompile[Person](`(?P<name>\S+) is (?P<age>\d+)`)
enc, _ := dec.Encoder()
s, _  := enc.Encode(Person{Name: "Alice", Age: 30})   // "Alice is 30"
back, _ := dec.One(s)                                  // Person{Name: "Alice", Age: 30}
```

### How it works

`(d *Decoder[T]).Encoder() (*Encoder[T], error)` parses the decoder's pattern (`regexp/syntax`, `syntax.Perl`) and walks the AST, inverting the **invertible subset** into an ordered encode plan:

- **Literal text** emitted verbatim (regexp escapes like `\.` already decoded by the parser).
- **Named captures** `(?P<name>…)` become field substitutions, resolved to struct fields with the **same field-mapping rules `Decoder` uses** (`regex:"name"` tag, else field name, case-insensitive; `regex:"-"` excluded). The sub-pattern is discarded — the field value fills the span.
- **Anchors / zero-width assertions** (`^ $ \A \z \b …`) match no text and are dropped.
- An **unnamed group** whose body is pure literal text is treated as that literal.

The value-formatting layer is reused unchanged: `encodeFieldValue`, the `RegexMarshaler` (`MarshalRegex() (string, error)`) extension point, the `encoding.TextMarshaler` fallback, the same conversion precedence as the decode side, and the `errors.As`-able `*EncodeError`.

### Fail-fast on non-invertible patterns

Any construct with no single string to emit **outside a named capture** — an alternation (`|`), a quantifier (`* + ? {n,m}`), a character class (`[...]`), an any-character wildcard (`.`), or an unnamed group with non-literal content — makes the pattern non-invertible, and `Encoder()` returns an error wrapping a new `ErrNotInvertible` sentinel that names the construct. Inside a named capture these are fine (the field value fills the group). Field-shape failures (a group mapping to no field, an unencodable field type) wrap `ErrInvalidStruct`, mirroring `Compile`. Construction is strict, so a successful `Encoder()` can only fail at `Encode` on a runtime value error (failing custom marshaler / nil pointer).

### What changed from the earlier design

This **supersedes the hand-written `{name}` template** approach discussed in the earlier design-note comment. The maintainer judged the template low-value: it is a second spec that must be kept in sync with the decode regex by hand, so you don't get the encoder "for free" from the decoder. Removed: `NewEncoder[T](template)`, `MustNewEncoder[T](template)`, `parseTemplate`, `Encoder.Template()`, and the `template` field. No hand-template escape hatch — it would reintroduce the two-spec drift.

**Deferred (documented future option):** re-matching each encoded value against its group's sub-pattern at `Encode` time — a perf/scope call, not done here.

Supported field types are unchanged: all scalar widths, `bool`, `time.Time` (RFC3339Nano by default or the `layout=` layout), `time.Duration`, single-level pointers, plus `RegexMarshaler` / `encoding.TextMarshaler`.

## Test plan

- [x] Derivation happy paths: field-name/tag/fold resolution, exact-beats-fold, literal-only, literal braces (now ordinary regex literals), dropped anchors, pure-literal unnamed group.
- [x] Derivation **rejection** of non-invertible patterns: alternation, `* + ? {n,m}` quantifiers, character class, bare `.` between groups, unnamed non-literal group, nested named-in-unnamed — each asserts `errors.Is(err, ErrNotInvertible)` and that the construct is named. Group-maps-to-no-field, excluded field, unsupported type → `ErrInvalidStruct`.
- [x] Type coverage: scalars, zero values, `time.Time` (canonical + `layout=`), `layout=` **round-trip symmetry**, `time.Duration`, pointers, `TextMarshaler`, custom `RegexMarshaler`; nil-pointer + nil-interface + marshaler errors (incl. `errors.Is`/`As` through `Unwrap`).
- [x] **Round-trip** contract: `Encode` → `Decoder.One` returns the original across scalars, sub-second `time.Time`, and custom-marshaler + pointer.
- [x] godoc `ExampleDecoder_Encoder` / `ExampleDecoder_Encoder_roundTrip`.
- [x] `encoder_bench_test.go` builds via the derived path; `BenchmarkEncode/simple` holds at **4 allocs/op, 88 B/op** (no regression vs the template implementation).
- [x] `gofmt`, `go vet`, `golangci-lint run` clean; `go test -race ./...` green; total coverage 96.7% (> 92% gate).

Refs #149


encoder.go
```diff
@@ -0,0 +1,492 @@
...
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
+		return strconv.FormatUint(field.Uint(), 10), nil
+	case reflect.Float32:
+		return strconv.FormatFloat(field.Float(), 'g', -1, 32), nil
+	case reflect.Float64:
+		return strconv.FormatFloat(field.Float(), 'g', -1, 64), nil
+	case reflect.Bool:
+		return strconv.FormatBool(field.Bool()), nil
+	default:
+		return "", fmt.Errorf("unsupported field type: %s", field.Kind())
```
Jecoms, line 598: `[dust]` A field whose static type is an interface implementing `encoding.TextMarshaler` (e.g. `F encoding.TextMarshaler`) passes `encodableType` at construction, but when its value is nil the `TextMarshaler` type-assertion yields `ok=false` and it falls through to this default, producing `unsupported field type: interface` rather than a nil-style `EncodeError` like the pointer path gives. Very rare edge; noting for awareness since nil handling is part of the encode contract. (encoder.go `encodeFieldValue` default case)

Jecoms (commented):
### Code review

**2 finding(s)** — 1 sand, 1 dust. See the threaded comments below.

Net-new public surface (`Encoder[T]`, `RegexMarshaler`, `*EncodeError`, the `{name}`/`{{`/`}}` template language) is well-shaped: it mirrors the decode side's precedence (`RegexMarshaler` → time special-cases → `TextMarshaler` → kind switch), `EncodeError.Error()` matches `DecodeError` (`field %s: %v`) and unwraps for `errors.Is`/`As`, construction is strict like `Compile`, and the round-trip/escaping/nil-pointer/time-layout contracts are documented and covered by tests (including sub-second RFC3339Nano fidelity and `errors.As` on `*EncodeError`). Both findings below are low-tier awareness items, not blockers.

CHANGELOG.md
Jecoms, line 1: `[sand]` This branch is 10 commits behind `main` and GitHub reports the PR as CONFLICTING (`mergeStateStatus: DIRTY`) — most likely the `### Added` section here, since several sibling entries landed on `main` since this branch forked. Rebase onto `main` and resolve before merge. (Verified via `gh pr view`/`gh api compare`.)

encoder.go
Jecoms: Fixed in 82f19fa. Added a nil-interface guard in `encodeFieldValue` right after the nil-pointer block:

```go
if field.Kind() == reflect.Interface && field.IsNil() {
    return "", fmt.Errorf("cannot encode nil interface of type %s", field.Type())
}
```

So a nil interface-typed field (e.g. `F encoding.TextMarshaler` or a `RegexMarshaler`-typed field holding no value) now returns a nil-style `*EncodeError` — mirroring the nil-pointer path — instead of falling through to the kind-switch default with `unsupported field type: interface`. A non-nil interface still passes through to the marshaler checks / kind switch on its dynamic value. Covered by a new `TestEncode_nilInterfaceErrors` asserting `errors.As` -> `*EncodeError` and that the message is the nil-style one rather than the fallthrough.

CHANGELOG.md
Jecoms: Done — rebased onto latest `main` (now at `650dfab9`, sentinel-errors #166) and resolved the `CHANGELOG.md` `### Added` conflict by hand. Kept both sides: `main`'s four new entries (required-group #148, `MissingNamedGroupsError` #151, `ErrInvalidPattern`/`ErrInvalidStruct` #152, and the fuller `DecodeError` #111 wording) plus this PR's `Encoder[T]` #149 entry, dropping the now-stale older duplicate `DecodeError` line. The rebase also surfaced a real build break: `main` changed `parseFieldTag` to return a 4th `required` value (#148), so `fieldCandidateName` was updated to discard it (encoding always emits the actual value, so `required` is irrelevant here). PR now reports `mergeable: MERGEABLE`. gofmt/vet/golangci-lint/tests/benchmarks all green.

Jecoms:
## Design note: why the custom `{name}` template, not `text/template`

We considered building the encode direction on `text/template` and are **keeping the custom mini-language** for the reversible core. Recording the rationale here so the decision is documented:

- **Reversibility is the contract, and `text/template` is one-directional.** Each `{name}` placeholder maps 1:1 to a `(?P<name>...)` capture group in the decode regex; `Encode` → `Unmarshal` round-trips *because* the names are shared between template and regex. `text/template` renders data→string with no inverse, and the decode side is regex-based — so it would serve only `Encode` and share zero structure with `Decoder`, losing the exact symmetry this feature exists to provide.
- **It's too powerful for a reversible grammar.** `{{if}}`, `{{range}}`, pipelines, and function calls have no inverse. Preserving round-trip would mean forbidding almost all of `text/template` — validating a whitelist against a much larger grammar, i.e. *more* work than the ~40-line `parseTemplate`, not less.
- **It doesn't address the hard problems.** The genuine risks (adjacent-placeholder decode boundaries, delimiter/escaping ambiguity) are decode-side constraints from the regex round-trip; render syntax doesn't help them.
- **Fit.** `{{.Field}}` collides with the `{{`/`}}` literal-escape already in use, and pulling in `text/template` (or a third-party reverse-template dep) is at odds with this being a small stdlib-`regexp`-focused library.

Where `text/template` *would* fit is a separate, **non-reversible** one-way "format a struct to a string" helper — out of scope for #149, and a candidate for its own issue if there's demand.

Jecoms:
## Redesign: derive the encoder from the decoder (supersedes the hand-template approach)

Reworked per review. The earlier design (see the design-note comment above) had the consumer hand-write a **separate `{name}` template** via `NewEncoder[T](template)`. That was correctly judged low-value: the template is a second spec that has to be kept in sync with the decode regex by hand, so you don't actually get the encoder "for free" from the decoder — which is the whole point of a round-trip inverse.

**What changed.** The primary API is now `(d *Decoder[T]).Encoder() (*Encoder[T], error)`. It derives the encode plan by parsing the decoder's **own compiled pattern** (`regexp/syntax`) and inverting the invertible subset of the AST:

- literal runs → emitted verbatim
- named captures `(?P<name>…)` → field substitutions (same field-mapping rules `Decoder` uses; sub-pattern discarded, the field value fills the span)
- anchors / zero-width assertions → dropped
- pure-literal unnamed groups → literals

The existing value-formatting layer (`encodeFieldValue`, `RegexMarshaler`, `encoding.TextMarshaler` fallback, `*EncodeError`, the nil-interface guard) is **reused unchanged** — only the source of the segment plan moved from a template parser to the AST walker.

**Fail-fast contract.** A construct with no single string to emit outside a named capture — alternation, quantifier, character class, `.`, or an unnamed non-literal group — makes the pattern non-invertible; `Encoder()` returns a construction error wrapping the new `ErrNotInvertible` sentinel (named construct in the message). Field-shape failures wrap `ErrInvalidStruct`, mirroring `Compile`'s strict validation.

**Removed** the two-spec surface entirely: `NewEncoder`/`MustNewEncoder`, `parseTemplate`, `Encoder.Template()`, and the `template` field. No hand-template escape hatch — keeping one would reintroduce exactly the drift this redesign removes.

**Not in scope (documented as a future option):** re-matching each encoded value against its group's sub-pattern at `Encode` time.

`Encode`'s hot path is untouched (still 4 allocs/op on the simple case). Full test/bench/doc/CHANGELOG updates included; `go test -race`, `go vet`, and `golangci-lint` all green.

Jecoms (commented):
### Code review

Re-review of the post-pivot approach (deriving `Encoder[T]` by inverting the decoder's compiled AST). The derivation is clean and well-documented, and the invertible/non-invertible classification, the encode/decode precedence mirror, and the RFC3339Nano time choice all check out.

**7 finding(s)** — 4 pebble, 2 sand, 1 dust. See the threaded comments below.

encoder.go
```diff
@@ -0,0 +1,588 @@
...
+	for i := range rt.NumField() {
+		sf := rt.Field(i)
+		if !sf.IsExported() {
+			continue
+		}
+		candidate, opts, skip := fieldCandidateName(sf)
+		if skip {
+			continue
+		}
+		if strings.EqualFold(candidate, name) {
```
Jecoms, line 383: `[pebble]` This fold pass matches an explicit `regex:` tag case-insensitively, but the decode side matches explicit tags **exactly** (`buildDecodePlan` folds only the untagged field-name fallback via `matchGroupName`; `subexpIndexes` is exact). So a case-mismatched explicit tag plus a `default=` silently diverges: `type P struct { ID string `regex:"ID,default=x"` }` with pattern `v=(?P<id>\S+)` — `Encode(P{ID:"foo"})` fold-binds group `id`→`ID` and emits `v=foo`, but `dec.One("v=foo")` maps `ID` to the absent group `ID` and returns the default `x`. Round-trip corruption with no error, contradicting the godoc's "resolves … exactly as `Decoder` resolves". Narrow trigger, but consider restricting the fold pass to the untagged (field-name) case to match the decoder. (decode side: decoder.go `matchGroupName`/`subexpIndexes`)

encoder.go
```diff
@@ -0,0 +1,588 @@
...
+//	    log.Printf("field %s (group %s) of type %s could not encode: %v", ee.Field, ee.Group, ee.Type, ee.Err)
+//	}
+//
+// Err holds the underlying cause (e.g. an error from a custom [RegexMarshaler]
+// or [encoding.TextMarshaler], or a nil-pointer field) and is reachable via
+// [errors.Is]/[errors.As] through Unwrap.
+type EncodeError struct {
+	// Field is the source struct field name.
+	Field string
+	// Group is the capture-group name the field resolved from — the field's
```
Jecoms: `[pebble]` `Group` is documented as the field's tag name "otherwise the field's own name", but `Encode` populates it from the **pattern's** capture-group name (`Group: seg.name`, `seg.name = re.Name`). Under a fold match they differ — pattern `(?P<AGE>\d+)` → field `Age` yields `Group == "AGE"`, not the field name. The decode-side `DecodeError.Group` godoc was tightened for exactly this imprecision in #142; mirror that wording here. (cf. unmarshal.go `DecodeError.Group`)

README.md
```diff
@@ -479,6 +535,7 @@ name, ok := regextra.FindNamed(re, "Alice 30", "name")  // "Alice", true
 - ✅ **Named group extraction** - Extract groups by name without index juggling
 - ✅ **Map-based access** - Get all named groups in one call
 - ✅ **Struct unmarshaling** - Type-safe extraction with automatic conversion
+- ✅ **Typed round-trip** - `Encoder[T]` renders a struct back to a string via a reversible template
```
Jecoms: `[pebble]` This Features bullet still says the encoder renders "via a reversible template", but the PR pivoted away from templates to deriving the encoder by inverting the decoder's compiled pattern. The CHANGELOG entry, the detailed README `Encoder()` section, and the `regextra.go` package doc were all updated to the new wording — this one bullet was missed and now contradicts the rest of the doc. (CONTRIBUTING.md: update README for public APIs)

encoder_test.go
```diff
@@ -0,0 +1,612 @@
...
+	e := mustEncoder[P](t, `(?P<age>\d+)`)
+	_, err := e.Encode(P{})
+	if err == nil {
+		t.Fatal("Encode of nil pointer field returned nil, want error")
+	}
+	var ee *rx.EncodeError
+	if !errors.As(err, &ee) {
+		t.Fatalf("error %v is not an *EncodeError", err)
+	}
+	if ee.Field != "Age" || ee.Group != "age" {
```
Jecoms, line 293: `[pebble]` The `EncodeError.Group` assertions only cover exact-name patterns (`"age"`, `"v"`), so nothing pins what `Group` carries on a fold match where the group name differs in case from the field name/tag — the same fold-case assertion #142 added on the decode side (decoder_test.go). Adding one would also lock down whatever the `Group`-godoc finding above resolves to. (CONTRIBUTING.md: test edge cases)

encoder_test.go
```diff
@@ -0,0 +1,612 @@
...
+	}
+}
+
+// ── Encode: custom RegexMarshaler ──────────────────────────────────────────────
+//
+// status (and its pointer-receiver UnmarshalRegex) is defined in
+// unmarshal_test.go; the value-receiver MarshalRegex below makes it a symmetric
+// round-trip type for the encode path.
+
+func (s status) MarshalRegex() (string, error) {
```
Jecoms, line 373: `[sand]` New exported interface `RegexMarshaler` has no runnable `ExampleRegexMarshaler`, though its decode-side mirror `RegexUnmarshaler` ships `ExampleRegexUnmarshaler` (unmarshal_test.go). It's shown in godoc prose and the README, so this is a consistency gap against the example-test convention rather than a hole. (CONTRIBUTING.md: "Add example tests for public APIs")

encoder.go
```diff
@@ -0,0 +1,588 @@
...
+//     text and are dropped.
+//   - An unnamed group whose body is pure literal text is treated as that literal.
+//
+// Any construct with no single string to emit — an alternation, a quantifier, a
+// character class, an any-character wildcard, or an unnamed group with
+// non-literal content — appearing outside a named capture group makes the pattern
+// non-invertible, and [Decoder.Encoder] fails fast with [ErrNotInvertible].
+//
+// [Decoder.Encoder] builds the plan once; [Encoder.Encode] walks it and
+// concatenates with a strings.Builder, running no per-call reflect of T's fields.
```
Jecoms: `[sand]` "running no per-call reflect of T's fields" reads as reflection-free, but `Encode` does reflect on every call (`reflect.New(...).Elem()`, `rv.Set(...)`, `rv.Field(...)`). The accurate claim is that it does no per-call *re-derivation of the field→group mapping* (that's cached in the plan). Consider a small wording tweak, e.g. "no per-call field-mapping reflection".

encoder.go
```diff
@@ -0,0 +1,588 @@
...
+	// Type is the source field's type, rendered (e.g. "int", "time.Time").
+	Type string
+	// Err is the underlying encode error.
+	Err error
+}
+
+// Error implements the error interface. The calling entrypoint prepends its own
+// `regextra.<Entrypoint>:` prefix when wrapping.
+func (e *EncodeError) Error() string {
+	return fmt.Sprintf("field %s: %v", e.Field, e.Err)
```
Jecoms, line 162: `[dust]` A directly-constructed `&EncodeError{}` renders a dangling `field : <nil>`. `Encode` always sets every field so this never occurs in practice, and it matches the accepted `DecodeError` direct-construction exposure noted in #159 — awareness only, likely no change.

encoder.go
Jecoms: Fixed in b063b97d. Restricted `resolveEncodeField`'s fold pass to untagged field names — an explicit `regex:` tag now binds to a group exactly, mirroring `subexpIndexes` on the decode side (only `matchGroupName`'s field-name fallback folds). Your `regex:"ID,default=x"` vs `(?P<id>…)` case now fails fast at `Encoder()` (`ErrInvalidStruct`, "no exported field") instead of fold-binding and corrupting the round-trip. Added `TestEncoder_explicitTagMatchesExactly` for it, and tightened the derivation godoc + CHANGELOG parenthetical (the tag is matched exactly, only the field name folds).

encoder.go
Jecoms: Fixed in b063b97d. Reworded `EncodeError.Group` to mirror the #142-tightened `DecodeError.Group`: "the field's `regex:"..."` tag name when set, otherwise the declared group whose name matches the field name — which may differ from the field name in case when the two matched via Unicode simple-fold."

README.md
Jecoms: Fixed in b063b97d. The Features bullet now reads "derived by inverting the decoder's own compiled pattern (no separate template to keep in sync)" — consistent with the CHANGELOG, the `Encoder()` README section, and the package doc.

encoder_test.go
Jecoms: Fixed in b063b97d. Added `TestEncode_foldMatchGroupName`: an untagged `Age` field against `(?P<AGE>\d+)` asserts `EncodeError.Group == "AGE"` (the declared group name, not the field name), locking down the fold-case Group contract the way #142 did on the decode side.

encoder_test.go
Jecoms: Fixed in b063b97d. Added `ExampleRegexMarshaler` mirroring `ExampleRegexUnmarshaler` (same `see Test… for a runnable demo" shape), closing the example-test consistency gap.

encoder.go
Jecoms: Fixed in b063b97d. Reworded to "It still reflects on the value each call to read fields, but does no per-call field-mapping reflection — the field↦group resolution is cached in the plan at construction."

encoder.go
Jecoms: Leaving as-is — awareness only, as you noted. `Encode` always populates every `EncodeError` field, so `field : <nil>` is unreachable via the public API, and this matches the accepted direct-construction exposure of `DecodeError` from #159. Not worth a guard.

Issue #172, opened 2026-07-02
Perf: small confirmed allocation wins across Encode/Compile/Find/Iter


Jecoms:
A benchmark-led optimization study (out of the #167 audit exploration) confirmed six small, safe, internal-only allocation wins across `Encode`, `Compile`/plan build, the `FindNamed` family, and `Decoder.Iter`, plus one clearly-marked opt-in judgment call on the Replace family. Each is an independent one-to-few-line change with an isolated, benchstat-confirmed delta and no exported API, error, or semantics change — filed as one umbrella since each item is S-effort.

- Related: #171 (the large win from the same study — decode-plan cache; subsumes part of items 2–3 below)
- Related: #167 / PR #170 — study origin; does **not** block either.
- Related: #154 (prior off-hot-path micro-opt pass; this extends it with newly confirmed items)

## Evidence / implementation notes

_All deltas from isolated serial re-measurement (benchstat, n=10–15, Apple M4 darwin arm64, p=0.000 unless noted); alloc deltas are deterministic and reproduced exactly. Every item passed adversarial verification (real-improvement / correctness / contract gates), `go vet`, and `go test -race -count=1 ./...` on its patched tree. The study scratchpad is temporary, so the load-bearing detail is carried here._

**1. `Encode`: `reflect.ValueOf(&v).Elem()` instead of `reflect.New` + `Set`** (`encoder.go` ~469–470)
Two-line swap to the addressable-parameter idiom; removes the fresh heap T plus the interface-boxing copy per call. `Compile` rejects non-struct T so `e.rtype == reflect.TypeOf(v)` always; the value stays addressable/settable so every pointer-receiver marshaler dispatch branch behaves identically (a mutating pointer-receiver-`MarshalRegex` probe passed under `-race` — mutations land in Encode's by-value param, caller untouched).
Delta: `Encode/simple` 146→126.8 ns, 4→3 allocs, 88→56 B (**-13.2% ns, -25% allocs, -36% B**); `Encode/withTime` -7.2% ns, 5→4 allocs; `Encode/manyFields` -7.6% ns, 5→4 allocs, 376→216 B (**-42.6% B**). Net: -1 alloc on every `Encode`, -7…-13% ns.

**2. `parseFieldTag` via `strings.Cut`** (`decoder.go` ~166)
[...]

PR #170, opened 2026-07-01
chore: public-contract accuracy & consistency audit before v2 (docs + error-type fixes)

Jecoms:
## Summary

Public-contract accuracy & consistency audit over the exported surface
(`regextra.go`, `decoder.go`, `unmarshal.go`, `encoder.go`, `README.md`) before
the v2.0.0 tag. Corrects documentation/comment drift against observed behavior,
plus one small code fix bringing `EncodeError.Error()` in line with its sibling
error types. No API break — the only code change is a nil-guard on a zero-value
`EncodeError` (error-message wording is explicitly non-breaking per README
§Stability).

### Documentation drift fixes (README)

1. **`NamedGroups` duplicate-name semantics** (README): "only the last match is returned" → the last *participating* occurrence wins, matching the godoc and #127.
2. **`Unmarshal` time layouts** (README): reorder to RFC3339Nano-first to match `timeLayouts` (`unmarshal.go`), fixing the contradiction with the Encoder section.
3. **`Decoder` vs `Unmarshal` allocations** (README + `regextra.go` package doc): refreshed to current measurements — `Decoder.One` 2 allocs, `Unmarshal` 11 allocs (were documented as 3/6) — and dropped the stale "half the allocations" claim.
4. **`Decoder.Iter` vs `UnmarshalAll` performance** (README): the "~37% faster / ~50% fewer allocations" claim is obsolete now that `UnmarshalAll` was optimized (#108/#41: 608→114 allocs). Restated as roughly comparable throughput on full iteration, with Iter's real advantage repositioned as streaming / lower peak memory / lazy early-break.
5. **Field-matching terminology**: unify decode-side "case-insensitive" and encode-side "Unicode simple-fold" — both are `strings.EqualFold` — on one precise phrase.
6. **Error-prefix contract** (README): document the intended exception that `Compile` / `Decoder.Encoder()` construction-time errors carry the bare `regextra:` sentinel prefix (they wrap `ErrInvalidPattern` / `ErrInvalidStruct` / `ErrNotInvertible`), not a `regextra.<Entrypoint>:` prefix.

### Error-taxonomy & supported-type godoc fixes (public entrypoints)

Follow-up commit correcting godoc that under- or over-stated the runtime contract:

[...]

PR #173, opened 2026-07-02
perf(unmarshal): cache the decode plan for free-function Unmarshal/UnmarshalAll

### LINKED ISSUE #171: Perf: cache the decode plan for free-function Unmarshal/UnmarshalAll
One-shot `Unmarshal`/`UnmarshalAll` rebuild the entire reflect decode plan on every call — exactly the work `Decoder[T]` caches away — so free-function callers pay the documented one-shot-vs-Decoder penalty for no structural reason. Caching the plan closes that gap almost entirely (`Unmarshal/simpleThreeField` drops from 567 ns / 11 allocs to 293 ns / 2 allocs) with no exported API, error, or semantics change.

- Related: #172 (umbrella of small confirmed wins from the same study; see non-additivity note below)
- Related: #167 / PR #170 — this issue came out of the #167 audit exploration and will invalidate doc numbers PR #170 just corrected; it does **not** block either.

## Evidence / implementation notes

_From a benchmark-led optimization study (Apple M4, darwin arm64; isolated serial re-measurement via benchstat, n=10–15, p=0.000 unless noted; alloc deltas deterministic; adversarially verified; all changes internal-only/contract-safe). The study scratchpad is temporary, so all load-bearing detail is carried here._

**Isolated measured deltas** (clean vs patched trees):

| Benchmark | before → after | ns/op | allocs/op | B/op |
|---|---|---|---|---|
| Unmarshal/simpleThreeField | 567.3→293.2 ns, 11→2 allocs, 507→96 B | **-48.3%** | **-81.8%** | -81.1% |
| Unmarshal/wideStruct | 3413→1071 ns, 48→2 allocs | **-68.6%** | **-95.8%** | -85.3% |
| UnmarshalAll/singleMatch | 676.3→396.6 ns, 14→5 allocs | -41.4% | -64.3% | -51.6% |
| UnmarshalAll/wideStruct | 69.9→67.1 µs, 103→57 allocs | -4.1% | -44.7% | -9.2% |
| UnmarshalAll/matches100 | ns flat (p=0.267) | 0% | -7.7% | -2.2% |

**Proposed mechanism.** Unexported package-level `sync.Map` keyed by `(pattern string, reflect.Type)` holding the immutable `[]fieldDecoder`. `buildDecodePlan(strict=false)` is provably pure in `(re.SubexpNames(), struct type)` — two `*regexp.Regexp` with equal `String()` cannot differ in `SubexpNames`. The plan is never mutated after build (`runDecodePlan`/`resolveGroupValue`/`setFieldValue` are read-only). 32-goroutine concurrent-first-use and cross-instance probes pass under `-race`. All three `buildDecodePlan` error returns are strict-guarded, so error behavior on the non-strict path is identical; the `Decoder` path is untouched.

**Decisions the implementer must surface at review:**

1. **Eviction policy** — process-lifetime no-eviction vs a bounded/LRU variant. Retention is small per (pattern, type) entry and does not pin the regexp, but workloads with unboundedly many distinct (pattern, type) pairs grow without bound. Ratify one explicitly.
2. **Doc updates in the same PR** — the `regextra.go` Performance section (~lines 63–66) and README ("Unmarshal does its reflect work per call"; Decoder "roughly half the time/allocations") go stale, **including the alloc numbers just corrected by PR #170** (Unmarshal = 11 allocs would become 2). Cross-check against #167 / PR #170 when landing.

**Rejected sibling — do not redo naively.** Precomputing `setFieldValue` type-dispatch predicates into the plan (`fieldDecoder.dispatch`) was REJECTED by the study: its four uncached `reflect.Implements` probes per field run on every free-function call, regressing one-shot `Unmarshal` by up to **+127% ns** (`timeFirstLayout` 933→2122 ns; `timeLastLayout` +113%; `simpleThreeField` +10% ns / +57% B/op) and `Compile` by +5–32% ns. It is only viable resubmitted with a `reflect.Type`-keyed dispatch cache so one-shot callers pay `reflect.Implements` once per type per process — re-verify one-shot + Compile benches if attempted.

**Effort / confidence:** M (cache + doc updates + eviction-policy decision) · high.

**Non-additivity:** this change subsumes the steady-state `Unmarshal` gains of the `parseFieldTag` and plan-pre-size items in #172; once the cache lands, their durable benefit is on `Compile`/`DeriveEncoder`/first-build (cache-miss) paths only. Do not sum percentages across the two issues.

## Implementation plan

_Added 2026-07-02 via /gh-analyze-issue. Review before implementing._

**Approach.** Unexported package-level `sync.Map` in `unmarshal.go` keyed by `{pattern string; typ reflect.Type}` (pattern = `re.String()`), holding the immutable `[]fieldDecoder`. Sound because non-strict `buildDecodePlan` depends on `re` only through `SubexpNames`/`SubexpIndex` (via `subexpIndexes`/`matchGroupName`) — pure functions of the pattern string — and the plan is never mutated after build. Recommend **process-lifetime, no eviction** (mirrors `encoding/json`'s type-keyed fieldCache; entries are small and don't pin the `*regexp.Regexp`) — ratify at review per decision 1 above. `Decoder`/`Compile` (strict) path untouched.

**Steps.**
1. Add a `getDecodePlan(rt, re)` helper: `Load` fast path; on miss `buildDecodePlan(rt, re, false)` then `LoadOrStore`, returning the stored value so racing first users share one canonical plan.
2. Route `Unmarshal` and `UnmarshalAll` through it; refresh their "uncached plan" comments (`unmarshal.go` ~206, ~273).
3. Tests in `unmarshal_test.go` (behavioral, per CONTRIBUTING's no-internals-pinning rule): interleaved distinct (pattern, type) pairs don't collide; two independently compiled equal-pattern regexps both decode correctly; ~32-goroutine concurrent first use (exercised under CI's `-race`).
4. Doc sweep: `regextra.go` Performance section (~lines 63–70); README `Decoder` section (the "eliminates the per-call reflect work" pitch and the ~540 ns / 11 allocs vs ~255 ns / 2 allocs comparison — **re-measure post-cache**, don't copy study numbers); stale comments in `decoder.go` ~145 ("built fresh per call") and ~270 ("every uncached … plan build").
5. `CHANGELOG.md`: Performance entry under `[Unreleased]` referencing #171.

**Files / surfaces touched.**
- `unmarshal.go` (cache + helper + comments), `unmarshal_test.go` (new tests), `decoder.go` (comments only), `regextra.go`, `README.md`, `CHANGELOG.md` (docs).

**Verification.**
- `go test -race ./...` green, including the new concurrent-first-use test.
- `benchstat` before/after per CONTRIBUTING (serial, low-resource runs): expect ≈−48% ns / −82% allocs on `Unmarshal/simpleThreeField`, ≈−69% / −96% on `wideStruct`; `Compile`/`Decoder`/`Encoder` benches flat. `BenchmarkCompilePlanOnly` (`bench_internal_test.go`) still covers the miss/build cost; `BenchmarkUnmarshal` now measures steady-state hits.

**Open questions / risks.**
- Eviction policy must be explicitly ratified at review (recommendation: unbounded; workloads with unboundedly many distinct patterns already pay per-call regexp compilation, so plan retention is second-order — document the growth behavior in the Performance section).
- Don't bundle the rejected `fieldDecoder.dispatch` precompute (see "Rejected sibling" above).
- Non-additive with #172's `parseFieldTag`/pre-size items — don't stack percentages across issues.

**References.**
- #172 (umbrella; non-additivity) · #167 / PR #170 (doc numbers this invalidates) · `encoding/json` fieldCache precedent for unbounded reflect-keyed caches.

Jecoms:
## Summary

- One-shot `Unmarshal`/`UnmarshalAll` rebuilt the entire reflect decode plan on every call — exactly the work `Decoder[T]` computes once at `Compile`. They now fetch the plan from an unexported process-lifetime `sync.Map` keyed by `(re.String(), reflect.Type)` via a new `getDecodePlan` helper (`Load` fast path; on miss `buildDecodePlan(strict=false)` then `LoadOrStore`, so racing first users converge on one canonical plan). The strict `Compile`/`Decoder` path is untouched.
- Soundness: the lenient `buildDecodePlan` reads the regexp only through `SubexpNames`/`SubexpIndex` (via `subexpIndexes`/`matchGroupName`) — pure functions of the pattern string, so two `*regexp.Regexp` with equal `String()` yield identical plans — and the plan is never mutated after build (`runDecodePlan` and friends are read-only). No public API, error, or semantics change.
- Doc sweep in the same PR: package-doc Performance section, README `Decoder` pitch + re-measured comparison numbers (invalidating the figures #170 just corrected — expected, see issue), stale "uncached/built fresh per call" comments in `unmarshal.go`/`decoder.go`, and a CHANGELOG `[Unreleased]` Performance entry.

## Decision for reviewers: eviction policy

The cache is **unbounded and process-lifetime — no eviction**. This is a deliberate choice that needs explicit ratification (issue #171, decision 1):

- Precedent: `encoding/json`'s type-keyed `fieldCache`, which is likewise never evicted.
- Entries are small (field indexes + parsed tag options) and do **not** retain the `*regexp.Regexp`.
- A workload decoding unboundedly many distinct `(pattern, type)` pairs grows the cache without bound — but such a workload already pays per-call regexp *compilation*, which dwarfs plan retention. The growth behavior is documented in the package-doc Performance section and the CHANGELOG entry.
- Alternative considered: a bounded/LRU variant — rejected as complexity without a demonstrated workload, per the plan's recommendation.

## Benchmarks (Apple M4, benchstat n=10, serial low-priority runs, p=0.000 unless noted)

| Benchmark | ns/op | allocs/op | B/op |
|---|---|---|---|
| Unmarshal/simpleThreeField | 558→290 (**−48.0%**) | 11→2 (**−81.8%**) | −81.1% |
| Unmarshal/wideStruct | 3549→1075 (**−69.7%**) | 48→2 (**−95.8%**) | −85.3% |
| UnmarshalAll/singleMatch | 675→393 (−41.8%) | 14→5 (−64.3%) | −51.6% |
| UnmarshalAll/wideStruct | −7.6% | −44.7% | −9.5% |
| Compile/* , DecoderOne/* | flat (±0–2% ns; identical allocs/B; drift also appears on the untouched `regexpCompileBaseline`) | — | — |

`BenchmarkCompilePlanOnly` still isolates the miss/build cost; `BenchmarkUnmarshal` now measures steady-state cache hits. Per issue #171: the rejected `fieldDecoder.dispatch` precompute is **not** bundled, and these gains subsume (are not additive with) #172's `parseFieldTag`/pre-size steady-state items.

## Test plan

- [x] `go test -race ./...` green, including three new behavioral tests in `unmarshal_test.go` (public-API only, no internals pinned): interleaved distinct (pattern, type) pairs don't cross-talk; two independently compiled equal-pattern regexps decode identically; 32-goroutine concurrent first use of a fresh pair (exercises first-build race under `-race`).
- [x] `gofmt -s -l`, `go vet`, `golangci-lint run` clean.
- [x] benchstat before/after (10 serial runs each, `nice -n 10`, no concurrent load) — table above.

Fixes #171


regextra.go
```diff
@@ -60,10 +60,18 @@ By use case:
 
 # Performance
 
-For one-shot extraction, [Unmarshal] does its reflect work per call. For repeated
-decode of the same shape (log parsers, request handlers, config readers), use
-[Compile] / [Decoder] — it caches the per-field plan and benchmarks at roughly
-half the time and far fewer allocations than [Unmarshal] on equivalent input.
+Every decode path caches its reflect work. [Unmarshal] / [UnmarshalAll] build
```
Jecoms: `[pebble]` "Every decode path caches its reflect work" overstates what is cached: decode still reflects on every call (`rv.Field` in `runDecodePlan`, `SetString`/`SetInt`/etc. in `setFieldValue`) — what's cached is the field→group decode plan. PR #168's review flagged this exact overstatement on the encode side (`[sand]`, "reads as reflection-free") and the agreed wording was "no per-call field-mapping reflection"; the decode-side sentence should use the same precision, e.g. "Every decode path caches its field-mapping reflect work." (source: PR #168 review thread, fixed there in b063b97d)

Jecoms (commented):
### Code review

**4 finding(s)** — 4 pebble. See the threaded comments below.

The core change checks out: the lenient `buildDecodePlan` reads the regexp only through `SubexpNames`/`SubexpIndex` (verified — `subexpIndexes` and `matchGroupName` are its only `re` touchpoints), `fieldDecoder` retains no `*regexp.Regexp` and no closures, the plan is read-only after build, and `LoadOrStore` converges racing first builders on one canonical plan. The new tests land in the correct sibling file and assert only the public contract (CONTRIBUTING.md §Testing). All four findings are doc/comment-accuracy leftovers from the PR's own stale-comment sweep — nothing blocks the code.

**Eviction policy — ratified.** The unbounded, process-lifetime cache is the right call: entries are small (field indexes + parsed tag options; the compiled regexp is not retained), the `encoding/json` `fieldCache` precedent is apt, and a bounded/LRU variant would be complexity without a demonstrated workload. The one way this differs from `json`'s cache — the key includes an arbitrary pattern string, so a fixed set of types with dynamically generated patterns still grows it where `json`'s pure type key would not — is exactly the workload that already pays per-call `regexp.Compile`, which dwarfs plan retention, and the growth behavior is disclosed in the package doc, `planCache` comment, and CHANGELOG. No follow-up needed.

1. `[pebble]` `unmarshal_bench_test.go`'s section-header cost models are now stale: the `BenchmarkUnmarshal` header still says each call runs "buildDecodePlan once (parseFieldTag + group-index resolution per field)" and the `BenchmarkUnmarshalAll` header says "buildDecodePlan once for the call" — after this PR, steady-state iterations hit the plan cache and never run `buildDecodePlan`, as the PR body itself notes ("BenchmarkUnmarshal now measures steady-state cache hits"). Worth folding into this PR's sweep since the file wasn't touched. (source: this PR's stale-comment sweep scope; PR body benchmark narrative)

   https://github.com/Jecoms/regextra/blob/db17238cbb25a7e5ea7b1a1a5ff9db6419958b70/unmarshal_bench_test.go#L264-L272

   https://github.com/Jecoms/regextra/blob/db17238cbb25a7e5ea7b1a1a5ff9db6419958b70/unmarshal_bench_test.go#L313-L320

decoder.go
Jecoms, line 1: `[pebble]` The exported `Decoder` godoc (decoder.go:32-34) still says the Decoder "eliminates the per-call reflect work that [Unmarshal] does on every invocation" — stale now that this PR makes Unmarshal cache its plan per (pattern, struct type) pair. The README's equivalent sentence was updated in this diff ("the Decoder's remaining edge is skipping the per-call cache lookup"), but this pkg.go.dev-visible doc was missed and now contradicts the updated package doc in the same PR. (source: this PR's stale-comment sweep scope; CONTRIBUTING.md §Documentation)

unmarshal.go
Jecoms, line 1: `[pebble]` `parseFieldTag`'s lazy-map rationale (unmarshal.go:405-409) still frames the saving as avoiding "a per-call, per-field empty-map allocation on the Unmarshal hot path" — after this PR, `parseFieldTag` runs only inside the at-most-once-per-(pattern, type) plan build, so the "per-call ... hot path" justification no longer holds (the optimization itself is still valid: it saves an allocation per cached plan entry). The directly analogous `resolveGroupName` cost comment in decoder.go was re-framed around the cache in this diff; this one was missed. (source: this PR's stale-comment sweep scope; cf. the resolveGroupName comment update in this diff)

regextra.go
Jecoms: Addressed in 940323c — reworded to "Every decode path caches its field-mapping reflect work," matching the precision agreed on the encode side in #168 (per-call conversion reflection via rv.Field/setFieldValue still runs; only the field→group plan is cached).

decoder.go
Jecoms: Addressed in 940323c — the Decoder godoc no longer claims Unmarshal reflects "on every invocation"; it now says Unmarshal caches its decode plan per (pattern, struct type) pair too, and frames the Decoder's remaining edge as skipping the per-call cache lookup plus the strict compile-time validation — same framing as the README sentence updated in this PR.

unmarshal.go
Jecoms: Addressed in 940323c — re-framed the lazy-map rationale around the cache: parseFieldTag runs only inside the at-most-once-per-(pattern, type) plan build, so the win is per retained plan entry (nil opts avoids an empty map held for the life of the process by every cached plan), not per call. Mirrors the resolveGroupName comment's cache framing.

Jecoms:
Pushed 940323c — thread responses are on the individual comments.

Re: the `unmarshal_bench_test.go` cost-model headers (summary finding #1, no thread): both section headers updated in 940323c. The `BenchmarkUnmarshal` header now describes the steady state — a `getDecodePlan` cache lookup + `runDecodePlan` per call, with `buildDecodePlan` running only on first use of a (pattern, struct type) pair — and the `BenchmarkUnmarshalAll` header gets the same cache-hit framing. No benchmark code changed, so no re-run needed.

Re: eviction-policy ratification — noted, thanks; leaving the unbounded process-lifetime cache as-is per the review.

PR #174, opened 2026-07-02
perf: small confirmed allocation wins across Encode/Compile/Find/Iter

### LINKED ISSUE #172: Perf: small confirmed allocation wins across Encode/Compile/Find/Iter

Jecoms:
## Summary

Lands the six benchstat-confirmed, internal-only allocation wins from #172 — one commit per item, in issue order. No exported API, error, or semantics change.

1. **`Encode`** reflects on its parameter's address (`reflect.ValueOf(&v).Elem()`) instead of `reflect.New` + `Set` — **−1 alloc on every Encode**, −7…−13% ns.
2. **`parseFieldTag`** walks tag pieces with `strings.Cut` instead of `strings.Split` — −1 alloc per tagged field per plan build. Post-#171 (plan cache), the durable benefit is `Compile`/`Encoder` derivation/cache-miss paths.
3. **`buildDecodePlan`** pre-sizes its plan slice to `rt.NumField()` (a strict upper bound). Nuance: an all-fields-skipped plan is now empty non-nil instead of nil — consumers only range/pass it, never nil-check.
4. **`FindNamed`** scans `re.SubexpNames()` inline instead of materializing `subexpIndexes`' slice (helper retained for its other callers) — **2→1 allocs/op, no-match paths 0 allocs**; duplicate-name last-participating-wins and `groupName==""` behavior preserved bug-for-bug.
5. **`Decoder.Iter`** reuses one zero-reset scratch `T` across matches, with a no-match early return — **a GC-pressure win, not a speed win**: `fullIteration` −48% allocs/op, −20% B/op, ns essentially flat. New regression tests pin the hazards: pointer-field aliasing across yields, optional-group leak across yields, re-ranging the same `Seq2`.
6. **`FindAllNamed`** collects occurrence indices in a stack-backed `[4]int` buffer (spills via append for >4 duplicate names, covered by a new test) — **−1 alloc/op everywhere**; known trade-off: borderline +1.2% ns at matches100, best for low-match-count callers.

Item 7 from the issue (`replaceNamed` `b.Grow`) is **excluded by default** per the implementation plan — it's an explicit maintainer opt-in (B/op −18…−26% vs +1.7–1.9% ns on largeManyMatches and capacity retention on shrink-heavy inputs). Happy to add it in this PR or a follow-up if wanted.

Includes a CHANGELOG entry under `[Unreleased]` → Performance.

## Type of change
- [ ] Feature (new public API or behavior)
- [ ] Fix (bug correction, no API change)
- [x] Chore / CI / docs / refactor (no behavior change) — `perf:` internal-only; no public API, error, or semantics change
- [ ] Breaking change (semver minor pre-1.0, major post-1.0)

## CHANGELOG

Under `## [Unreleased]` → `### Performance`:

> - **Six small confirmed allocation wins across `Encode`, `Compile`/plan build, `FindNamed`/`FindAllNamed`, and `Decoder.Iter`.** Each is an internal one-to-few-line change with a benchstat-confirmed delta and no public API, error, or semantics change: (1) `Encode` reflects on its parameter's address instead of `reflect.New` + `Set` — −1 alloc on every call, −7…−13% ns; (2) `parseFieldTag` walks tag pieces with `strings.Cut` instead of `strings.Split` — −1 alloc per tagged field per plan build (post-#171 this benefits `Compile`/`Encoder` derivation/cache-miss paths); (3) `buildDecodePlan` pre-sizes its plan slice to `NumField()` — `Compile/fields20` −14% ns, −42% B/op; (4) `FindNamed` scans `SubexpNames()` inline instead of materializing an index slice — 2→1 allocs/op, no-match paths 0 allocs; (5) `Decoder.Iter` reuses one zero-reset scratch value across matches instead of allocating a fresh `T` per match — a GC-pressure win (`fullIteration` −48% allocs/op, −20% B/op, ns flat), with regression tests pinning yield-aliasing safety; (6) `FindAllNamed` collects occurrence indices in a stack-backed buffer — −1 alloc/op everywhere (borderline +1.2% ns at 100 matches, best for low-match-count callers). ([#172](https://github.com/Jecoms/regextra/issues/172))

## Test plan

- [x] `gofmt -s -l .` clean, `go vet ./...` clean, `golangci-lint run` 0 issues
- [x] `go test -race -count=1 ./...` green (includes the three new Iter aliasing/leak/reuse regression tests and the FindAllNamed spill-path test)
- [x] Alloc deltas re-confirmed vs trunk (serial, low-resource, `-benchmem`, benchstat): deterministic allocs/op match the issue's table exactly — Encode 4→3/5→4, DecoderIter/fullIteration 205→106 (noMatch stays 0), FindNamed 2→1 (no-match 1→0), FindAllNamed −1 everywhere, CompilePlanOnly 10→5 and Compile/fields20 62→37 allocs (items 2+3 stacked)

Fixes #172

## Summary by CodeRabbit

* **Performance**
  * Improved encoding, decoding, and named match lookup performance with fewer temporary allocations.
  * Reduced work during repeated decoding and regex match collection for better efficiency on larger inputs.

* **Bug Fixes**
  * Preserved existing behavior while making named capture handling more consistent when a name appears multiple times.
  * Tag parsing remains unchanged functionally, with more efficient processing behind the scenes.


Jecoms (commented):
### Code review

**4 finding(s)** — 1 pebble, 3 dust. See the threaded comments below.

Overall: the six allocation wins check out. Behavior preservation was verified in depth — the Iter scratch-reuse zero-reset (pointer-field aliasing, optional-group leak, Seq2 reuse), the FindNamed inline scan (duplicate-name last-participating-wins, non-participating groups), the FindAllNamed `[4]int` spill path, the Encode addressability swap (pointer-receiver marshaler dispatch), the parseFieldTag `strings.Cut` walk (trailing/doubled commas, lone flags, map size hint), and the empty-non-nil plan slice all match trunk semantics; git history shows no prior fix being reverted, and the new regression tests pin the right hazards. The only substantive finding is a wording error in a newly added invariant comment.

1. `[dust]` The PR body omits the "## Type of change" section from the PR template; the classification is unambiguous from the `perf:` title, so this is surfaced for awareness only. (.github/PULL_REQUEST_TEMPLATE.md)

2. `[dust]` The PR body omits the template's "## CHANGELOG" section. The CHANGELOG.md entry itself is present and well-formed in the diff, and the template scopes the paste-into-body instruction to "features and fixes", so a perf PR is arguably outside it. (.github/PULL_REQUEST_TEMPLATE.md)

unmarshal.go
```diff
@@ -404,13 +410,15 @@ func parseFieldTag(field reflect.StructField) (name string, opts map[string]stri
...
-		// inside the at-most-once-per-(pattern, type) plan build, so the win
-		// is not per call but per plan entry: nil opts avoids an empty map
-		// retained for the life of the process by every cached plan. Consumers
-		// already treat nil opts as "no options" (nil-map reads are zero-value).
+		// matching the no-options case (no comma in the tag). parseFieldTag
+		// runs only inside the at-most-once-per-(pattern, type) plan build, so
+		// the win is not per call but per plan entry: nil opts avoids an empty
+		// map retained for the life of the process by every cached plan.
+		// Consumers already treat nil opts as "no options" (nil-map reads are
+		// zero-value). The size hint (comma count = piece count) matches the
```
Jecoms: `[pebble]` The parenthetical "comma count = piece count" in this new comment is wrong as written: a tag with n commas splits into n+1 pieces, so the comma count equals the option-piece count (`len(parts)-1`), not the total piece count. The size hint itself is correct — only the sentence needs rewording, e.g. "comma count = option-piece count". (source: the comment's claim vs `strings.Split`/`strings.Cut` semantics; `strings.Count("name,a=1,b=2", ",")` = 2, pieces = 3)

decoder.go
```diff
@@ -161,7 +161,9 @@ func compileDecoder[T any](pattern string, re *regexp.Regexp) (*Decoder[T], erro
 // setFieldValue. This preserves Unmarshal's historical best-effort behavior, so
 // buildDecodePlan never returns a non-nil error when strict=false.
 func buildDecodePlan(rt reflect.Type, re *regexp.Regexp, strict bool) ([]fieldDecoder, error) {
-	var fields []fieldDecoder
+	// NumField is a strict upper bound on plan entries (the loop only skips),
+	// so one right-sized allocation replaces append's grow-and-copy chain.
+	fields := make([]fieldDecoder, 0, rt.NumField())
```
Jecoms, line 166: `[dust]` The all-fields-skipped plan is now empty non-nil instead of nil — documented in the PR body but not pinned by a test (no existing test covers a struct where every field is skipped). Nothing observable changes through the public API since consumers only range the plan, but a cheap compile+decode test on an all-skipped struct would lock the invariant, matching the pattern prior reviews asked for when decode-plan internals change. (source: prior review pattern, PR #143)

unmarshal.go
Jecoms: Addressed in 02d907ac798d — reworded the parenthetical to "comma count = option-piece count, since n commas split into n+1 pieces and the first piece is the name", matching the old `len(parts)-1`. Comment-only; the size hint itself was already correct.

decoder.go
Jecoms: Addressed in 02d907ac798d — added `TestDecoder_allFieldsSkippedDecodesToZeroValue`: a struct whose every field is excluded (`regex:"-"` + an unexported field) must still compile, and `One` on a matching input returns the zero value. This pins the empty non-nil plan invariant from the pre-sizing change.

Jecoms:
Pushed 02d907ac798d — responses to the two inline findings are on their threads.

Re: the two review-level template findings — the PR description now includes the "## Type of change" section (Chore/no behavior change checked, with a perf note) and the "## CHANGELOG" section pasting the `[Unreleased]` → Performance entry from the diff.

